PR #3, opened 2014-12-31
Add redundant class to h1s matching package description

bakkerthehacker:


zeke:
Thanks for the patch, but I botched merging it. :see_no_evil: 

Landed in https://github.com/zeke/marky-markdown/commit/ea73a2a59a5c344063f4012edf3112e44f53b643

PR #6, opened 2015-01-06
Use markdown-it replace marked to resolve performance

### LINKED ISSUE #5: Consider switch to `markdown-it` parser
In https://github.com/npm/newww/issues/487#issuecomment-68662039 @KenanY recommended to forward issue to your repo.

Consider switch to [markdown-it](https://github.com/markdown-it/markdown-it) - much more safe than current one.

fengmk2:
Fixes #5

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -1,9 +1,42 @@
 var hljs = require("highlight.js")
-var marked = module.exports = require("marked")
+var MarkdownIt = require("markdown-it")
 
-marked.setOptions({
+var md = new MarkdownIt({
+  html: true,
   langPrefix: "hljs ",
   highlight: function(code) {
     return hljs.highlightAuto(code).value;
   }
 })
+
+module.exports = function render(markdown) {
+  // return require('marked')(markdown)
+  return md.render(markdown)
+}
+
+// override default heading_open render, add dom id attribute
+// # h1 => <h1 id="h1">h1</h1>
+md.renderer.rules.heading_open = function (tokens, idx, options, env, self) {
+  // default impl:
+  // ```js
+  // return '<h' + tokens[idx].hLevel + '>';
+  // ```
+  var hLevel = tokens[idx].hLevel
+  var nextToken = tokens[idx + 1]
+  var raw = nextToken && nextToken.content
+  if (!raw) {
+    return "<h" + hLevel + ">"
+  }
+
+  return "<h" + hLevel + " id=\"" + id(raw) + "\">"
+}
+
+function id(str) {
+  // replace " " and all punctuation characters to "-"
+  str = str.toLowerCase().replace(/[\s\]\[\!\"\#\$\%\&\'\(\)\*\+\,\.\/\:\;\<\=\>\?\@\\\^\_\`\{\|\}\~\-]+/g, "-")
+  try {
+    return encodeURIComponent(str)
+  } catch (e) {
+    return str.replace(/[^\w]+/g, "-")
+  }
+}
--- a/test/index.js
+++ b/test/index.js
@@ -510,3 +510,26 @@ describe("real readmes in the wild", function() {
 
   })
 })
+
+describe("performance", function() {
+
+  it("successfully parses big markdown", function() {
+    var $ = marky(fixtures.performance)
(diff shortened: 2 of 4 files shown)


lib/render.js
```diff
@@ -1,9 +1,43 @@
...
     return hljs.highlightAuto(code).value;
   }
 })
+
+module.exports = function render(markdown) {
+  // return require('marked')(markdown)
+  return md.render(markdown)
+}
+
+var defaultRender = md.renderer.rules.heading_open
```
puzrin: I think, here is no need to keep default renderer, because it will be called only for empty headers. Header always generate this sequeance: `header_open` + `inline` + `header_close`, and i can't imagine situation when this can be changed.

lib/render.js
```diff
@@ -1,9 +1,43 @@
...
+  if (!raw) {
+    return defaultRender(tokens, idx, options, env, self)
+  }
+
+  var hLevel = tokens[idx].hLevel
+  return "<h" + hLevel + " id=\"" + id(raw) + "\">"
+}
+
+function id(str) {
+  str = str.toLowerCase().replace(/[\s`\.\'\"\(\)\[\]\/\\]+/g, '-')
```
puzrin: It seems, github strips dots, instead of replace to dash

lib/render.js
fengmk2: ok

test/index.js
```diff
@@ -510,3 +510,25 @@ describe("real readmes in the wild", function() {
...
+describe("special characters", function() {
+
+  it("successfully parses h1 contains \"\\udeff\"", function() {
+    var $ = marky("# foo\udeffbar")
+    assert.equal($.html(), "<h1 id=\"foo-bar\" class=\"deep-link\"><a href=\"#foo-bar\">foo&#xDEFF;bar</a></h1>\n")
+  })
+
+  it("successfully parses h1 contains ^\\w", function() {
+    var $ = marky("# 中文a\'b\"c`d|e,f.g?h&i/j\\k")
+    assert.equal($.html(), "<h1 id=\"%E4%B8%AD%E6%96%87a-b-c-d%7Ce%2Cf-g%3Fh%26i-j-k\" class=\"deep-link\"><a href=\"#%E4%B8%AD%E6%96%87a-b-c-d%7Ce%2Cf-g%3Fh%26i-j-k\">&#x4E2D;&#x6587;a&apos;b&quot;c`d|e,f.g?h&amp;i/j\\k</a></h1>\n")
```
fengmk2: test on github:

# 中文a\'b\"c`d|e,f.g?h&i/j\\k

lib/render.js
puzrin: https://github.com/markdown-it/markdown-it/blob/master/CHANGELOG.md#300--2015-01-04

See anchor, in my case dots are stripped.

lib/render.js
fengmk2: remove now.

lib/render.js
```diff
@@ -1,9 +1,42 @@
...
+  if (!raw) {
+    return "<h" + hLevel + ">"
+  }
+
+  return "<h" + hLevel + " id=\"" + id(raw) + "\">"
+}
+
+function id(str) {
+  // replace " " and all punctuation characters to "-"
+  str = str.toLowerCase().replace(/[\s\]\[\!\"\#\$\%\&\'\(\)\*\+\,\.\/\:\;\<\=\>\?\@\\\^\_\`\{\|\}\~\-]+/g, "-")
```
fengmk2: @puzrin I dont want to strip dots, because marked replace dots to "-". These can keep compatibility.

puzrin: As you wish. I don't know correct behaviour. Only noticed unfamiliar thing.

I think, here it enougth to do marked-like logic for now, and place note to makdtownun-it tracker, where we discuss ankors issues.

What do you think?

/cc @zeke

bcoe:
@fengmk2 thanks for the contribution!

We ended up combining your changes with this pull: https://github.com/npm/marky-markdown/pull/15

marky-markdown should now exhibit the behavior you were looking for.

fengmk2:
Cool!

PR #13, opened 2015-01-26
Add missing squiggles to tests

psalaets:
Without squiggle the assert checks "string not at beginning" instead of "string is present"


zeke:
thanks, @psalaets!

PR #15, opened 2015-01-27
[WIP] start using markdown-it and highlights

zeke:
This PR is doing two things at once:

- replace marked with markdown-it for better performance and stability
- replace pygments-bundled with highlights for, well, better performance and stability

@bcoe if you end up working on this, notice that I started a span whitelist, but it's still missing some of the tokens: https://github.com/atom/highlights/issues/13

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,9 +44,6 @@ var marky = module.exports = function(markdown, options, callback) {
   log("Convert HTML frontmatter into meta tags")
   html = frontmatter(markdown)
 
-  log("Remove HTML comments")
-  html = comments(html)
-
   log("Parse markdown into HTML and add syntax highlighting")
   render(html, options, function(err, output) {
     html = output
--- a/lib/render.js
+++ b/lib/render.js
@@ -1,5 +1,8 @@
-var highlighter = new (require("highlights"))()
 var _ = require("lodash")
+var escapeHtml = require('escape-html')
+var highlighter = new (require("highlights"))()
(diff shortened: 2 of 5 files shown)


bcoe:
@zeke this looks good to me, with the caveats:

* I'd love to dig into the bug I was seeing with `<!--` when I get home.
* I'd like to re-run the benchmarking tool on all the packages, to vet the two new libraries.

bcoe:
@zeke finished running [the benchmarker](https://github.com/bcoe/benchmark-markdown). Some notes:

* after today's various patches, there are no longer any exceptions that occur rendering any npm READMEs.
* `highlights` and `markdown-it` run blazing fast:
  * average rendering speed was 16ms:
    * the slowest package to render was `npm-compact`, at 4.4s. This package appears to be an outlier, containing a ridiculously large JSON blob in it.
    * the only other package that took longer than 2s to render was `jsr-constants`, again a file with a huge amount of JSON.
* I've added auto-headline-id-generation back, but prefix the ids with `user-content-`, and sanitize the text thoroughly.
  * my thinking is that we can have some JavaScript on the client-side that handles inline links, and attempts both the `user-content` anchor and the original anchor. this allows us to support both manually inserted `<a name="" />` and the auto-generated ids.
* there's a corresponding pull for `newww` which adds the syntax highlighting styles from Atom:
  * https://github.com/npm/newww/pull/582

screen shot 2015-01-26 at 10 29 50 pm

PR #23, opened 2015-02-16
Update CLI to use the synchronous API

eush77:
Apparently CLI is out-of-date now. This is a quick fix.

Also, shouldn't the main module throw errors rather that returning them?


zeke:
Thanks!

I should probably wraps some tests around the CLI with http://npm.im/nixt -- I'll take a look at this and merge when I get into the office today. :)

zeke:
Landed in 4.3.0

Thanks again, @eush77

PR #25, opened 2015-02-17
added haxe, nginx, and erb

bcoe:
* added haxe and nginx grammars.
* changed how mappings work, adding support for erb.


zeke:
:lollipop: 

Landed in https://github.com/npm/marky-markdown/releases/tag/v4.3.0

PR #27, opened 2015-02-18
Update sanitize-html to 1.6.1

aearly:
This fixes a bug evidenced in it's own changelog on npm: https://www.npmjs.com/package/sanitize-html#changelog -- a code block that just contains `undefined` will be squashed to an empty string.


zeke:
Thanks!

PR #32, opened 2015-02-19
Don't assert() outside unit tests

eush77:
[`assert`](http://nodejs.org/api/assert.html) is not supposed to be used outside unit tests.

> This module is used for writing unit tests for your applications

Using this module in regular code adds an extra dependency that needs to be shimmed in other environments ([browserify shim](https://github.com/defunctzombie/commonjs-assert/blob/master/assert.js) is pretty huge) and most of its methods won't make sense anyway (`throws`, `deepEqual`, `fail`).

Simple `throw` statement here is more than enough.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,3 @@
-var assert      = require("assert")
 var cheerio     = require("cheerio")
 var defaults    = require("lodash").defaults
 var comments    = require("./lib/comments")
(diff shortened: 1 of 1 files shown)


zeke:
marky-markdown is unlikely to ever fully work in the browser due to the C dependencies in `highlights`, but I appreciate the sentiment here and generally think it wise to consider the browserify implications. :)

PR #38, opened 2015-04-11
[WIP] Add support for IDs in headings with links or images

wooorm:
The reason for this is that I have lots of docs up on Readme’s, with headings such as `foo.bar(Baz)`, where `foo` links to the docs of the object, and `Baz` links to docs for that class.

It does change some things in the UI of npm though!

*   This adds support for links and images in headings
    by moving the anchor element, which previously wrapped
    the whole heading’s contents, before the contents
    (thus being empty).  In addition, this anchor
    element receives a new class name (`"deep-anchor"`,
    in analogy with `"deep-link"`) to make it stand out
    from other links.

*   The addition does change the UI by no longer making
    headings clickable. Where previously `Foo` was
    completely clickable, now only the `#` preceding the
    heading is clickable (on hover).

*   It does require a change in the styling of npm, e.g.:

    ```css
    .markdown > h2.deep-link a:hover::before {...}
    .markdown > h2.deep-link a::before {...}
    ```

    ...should/might be:

    ```css
    .markdown > h2.deep-link:hover a.deep-anchor::before {...}
    .markdown > h2.deep-link a.deep-anchor::before {...}
    ```

    I tested in out in my web inspecter and things worked
    quite well :)

### DIFF
--- a/lib/headings.js
+++ b/lib/headings.js
@@ -10,7 +10,7 @@ var headings = module.exports = function($, options) {
   $("h1,h2,h3,h4,h5,h6").each(function(i, elem) {
 
     // Bail if heading already contains a hyperlink
-    if ($(this).find("a").length) return
+    if ($(this).find("a.deep-anchor").length) return
 
     // Generate an ID based on the heading's innerHTML
     if (!$(this).attr("id")) {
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
given the changes required of the npm site right now, this can't be merged, but will keep in mind for future versions. thanks!

ashleygwilliams:
closing in favor of https://github.com/npm/marky-markdown/pull/225

PR #45, opened 2015-04-24
Allow `colspan`/`rowspan` on table cells

dominykas:
Because I have a module, that has a complex table in it...

By the way, any reason the styling of tables on npm site is the way it is? Github centers (or, rather, does not reset to `left`) the `th`. Also, borders do help... 

Compare:
https://www.npmjs.com/package/gampee#params
vs
https://github.com/insidewarehouse/gampee#params


zeke:
Thanks, @dominykas. This landed in 5.3.0.

I agree that the website's table styles need some help. Patches welcome!

PR #53, opened 2015-09-04
"Memory leak" fixes

soldair:
[highlights](https://npmjs.org/package/highlights) uses [first-mate](https://npmjs.org/package/first-mate) generated rules. They take a lot of memory. if you run this module in a server environment handling many readmes for many languages over a long time these rules can take well over a GB of heapUsed.

this "fixes" this by cleaning up highlighter.registry.grammars rules that have been inactive for 2 seconds using [property-tll](https://npmjs.org/package/property-ttl). In batch testing generating the rules doesn't seem to take more than a few ms so the "cold cache" case is not going to cause much event loop lag.

This does not expose the ability to configure the ttl at this time. It's a little awkward as its nested in lib/render.js and highlights itself is never exposed which is also an issue, but for another pr.

All timers created to manage this are unrefed so this doesn't cause processes to hang unexpectedly.


aredridel:
Nice. Build looks good, failure is unrelated.

soldair:
thanks!

Issue #59, opened 2015-10-18
Support parsing of markdown syntax for emojis


patriciarealini:
I am working on the following issue for [npm/newww](https://github.com/npm/newww).

[<img width="780" alt="screen shot 2015-10-18 at 11 06 30 am" src="https://cloud.githubusercontent.com/assets/7608497/10564608/57f22812-7588-11e5-8bd7-ea3406e41f78.png">](https://github.com/npm/newww/issues/463)

I was thinking I would approach this issue by requiring [Markdown-It-Emoji](https://github.com/markdown-it/markdown-it-emoji) in [marky-markdown](https://github.com/npm/marky-markdown) (which already requires [Markdown-It](https://github.com/markdown-it/markdown-it)).

ashleygwilliams:
hey! this sounds p cool. could you open a WIP PR so i can get a sense of how you would implement it? thanks for helping!

patriciarealini:
https://github.com/npm/marky-markdown/pull/87

PR #60, opened 2015-10-18
[WIP] Support parsing of markdown syntax for emojis 

patriciarealini:
install & require [markdown-it-emoji](https://github.com/markdown-it/markdown-it-emoji) for Issue #59

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,17 @@
-sudo: false
 language: node_js
-node_js:
-  - "0.12"
-  - "0.10"
-  - iojs
+
 notifications:
   email: false
+
+node_js:
+  - 0.10
+  - 0.12
+  - iojs
+  - 4
+  - 5
+
+before_install:
+  - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq -y g++-4.8
+  - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 50
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # marky-markdown
 
+[![Build Status](https://travis-ci.org/npm/marky-markdown.svg?branch=master)](https://travis-ci.org/npm/marky-markdown)
+[![Code Climate](https://codeclimate.com/github/npm/marky-markdown/badges/gpa.svg)](https://codeclimate.com/github/npm/marky-markdown)
+[![Dependency Status](https://david-dm.org/npm/marky-markdown.svg)](https://david-dm.org/npm/marky-markdown)
+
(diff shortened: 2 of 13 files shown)


zeke:
:clap: :clap: :clap:

patriciarealini:
thanks @zeke!

patriciarealini:
The only thing, i think, i need to do to complete this PR is write the test. But i could use a little help understanding the formatting of the tests. 

@zeke I noticed that many of the commits on the `test/index.js` file were made by you, do you think you could walk me through the reasoning for the current format of the tests? I'm a little lost as to why they only compare the `.indexOf` the markdown to the `.length` of the DOM element that _should_ take it's place. I tried to mirror this test & it failed. (Which is totes fine! I would rather learn how to write better tests than just write one that passes cause it's the same as all the others. but i also wanted to stay as true as possible to the current state of affairs. trying to keep this contribution as seamless as possible...)

this is the direction i was going in...
```
    it('replaces markdown syntax for emoji with unicode for the emoji', function () {
      assert(~fixtures.github.indexOf(':sparkles:'))
      assert($('&#x2747;').length)
    })
```
the result i get is...
```
     SyntaxError: Unmatched selector: &#x2747;
```
even though I have tested the output in HTML for `:sparkles:` and it is `&#x2447;` (in alignment with the emoji unicode directory that markdown-it-emoji uses) please halp.

soldair:
oh! i know the `$(" ")` only takes a css selector like an id `#id` or `.class` etc. 
you will have to compare the result text something like this i think
```
assert(~$.html().indexOf('&#x2747;'))
```

zeke:
@soldair is on the right track there. To answer your question, @patriciarealini:

```js
it('removes classnames from elements', function () {
  assert(~fixtures.dirty.indexOf('class="xxx"'))
  assert(!$('.xxx').length)
})
```

The above code is ensuring that the given substring is **present in the fixture, but not present in the cheerio-generated DOM tree**. The `~` is the magical and not-very-intuitive [bitwise NOT](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators?redirectlocale=en-US&redirectslug=JavaScript%2FReference%2FOperators%2FBitwise_Operators#Bitwise_NOT). It would work just as well (and be more clear) to do this:

```js
assert(fixtures.dirty.indexOf('class="xxx"') > -1)
```

or...

```js
const contains = require('lodash').contains
assert(contains(fixtures.dirty, 'class="xxx"')
```

ashleygwilliams:
hey @patriciarealini i'm thinking i might grab this PR and see it though so we can merge it. i'd like to cut a new minor version today with the dep updates i did and this! thoughts?

patriciarealini:
ya sounds good! @ashleygwilliams

ashleygwilliams:
closing in favor of #87

PR #65, opened 2015-11-16
add support for ~~strikethrough~~

zeke:
The markdown-it parser supports [GFM-style strikethrough](https://github.com/markdown-it/markdown-it#syntax-extensions), but the `<s>` element was not allowed by marky-markdown's sanitizer. This PR adds it to the whitelist.

> The HTML Strikethrough Element (`<s>`) renders text with a strikethrough, or a line through it. Use the `<s>` element to represent things that are no longer relevant or no longer accurate.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element

Fixes https://github.com/npm/newww/issues/1065; cc @jamestalmage

### DIFF
--- a/lib/sanitize.js
+++ b/lib/sanitize.js
@@ -5,7 +5,7 @@ var sanitizer = module.exports = function (html) {
 
 sanitizer.config = {
   allowedTags: sanitizeHtml.defaults.allowedTags.concat([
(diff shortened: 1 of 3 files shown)


jamestalmage:
:clap: :+1:

soldair:
awesome. looks great

PR #77, opened 2015-12-03
an exception was being thrown if an image-style tag did not contain a URL

bcoe:
An exception was being thrown on the page:

https://staging.npmjs.com/package/gekoq

Because of the line:

```gfm
![Methods](#methods)
```


ashleygwilliams:
:bow: lgtm!

Issue #55, opened 2015-09-11
superscript tags are stripped out


scurker:
This is related specifically to superscript tags, but could be related to any other relative safe html tag. I ran into this in the documentation of currency.js ([Github](/scurker/currency.js) vs [npm](https://www.npmjs.com/package/currency.js)) where a superscript tag is being stripped out which leads to a fundamental difference in how the documentation is read.

**Github:**
![github](https://cloud.githubusercontent.com/assets/1062039/9804702/26c2c0d0-57f6-11e5-8927-fe1dfd8fccb5.png)

**npm:**
![npm](https://cloud.githubusercontent.com/assets/1062039/9804724/8a7fd7d4-57f6-11e5-8b04-18d4ea53adcf.png)

This seems like a relative safe tag to include, and it would be nice if npm could also match [Github's whitelist of tags](https://github.com/jch/html-pipeline/blob/2a55955b5da482624df95ff5cf88ba7ff522e97a/lib/html/pipeline/sanitization_filter.rb#L112) in order to have similar formats for readme documentations that are listed in both places.

ashleygwilliams:
fixed by #83

PR #83, opened 2015-12-28
Whitelist the ins, del, sub, and sup HTML elements

revin:
Per issue #55, this update white lists a few HTML elements from [GitHub's whitelist](https://github.com/jch/html-pipeline/blob/2a55955b5da482624df95ff5cf88ba7ff522e97a/lib/html/pipeline/sanitization_filter.rb#L112). Most of the elements in that list are already allowed, so this just adds `del`, `ins`, `sub`, and `sup` (and it alphabetizes the array for `allowedTags`). There's also a simple test fixture for each.


ashleygwilliams:
tests are passing. looks good! thank you so much!

Issue #64, opened 2015-11-01
diff / patch syntax highlighting


dsernst:
Reposted from https://github.com/npm/newww/issues/1432

----------------

GitHub supports `diff` syntax highlighting:

```diff
old line
old line
+ insert line
old line
- remove line
old line
```

But [npmjs.com](http://npmjs.com) does not. :cry:

ashleygwilliams:
fixed in #84

dsernst:
:sunglasses:

PR #84, opened 2015-12-29
Support diff syntax highlighting

revin:
Yaks be shaved, this should take care of #64:

* The stock [highlights](https://github.com/atom/highlights) package we use for syntax highlighting doesn't include a grammar for diffs. Its existing grammars are mostly automatically converted from TextMate grammars, and TM _does_ have a diff one, so I converted it myself and published it to NPM as [atom-language-diff](https://www.npmjs.com/package/atom-language-diff), then added it as a dependency here.
* I discovered that the HTML sanitizer was wrecking the generated output. It turns out that the [highlights-tokens](https://www.npmjs.com/package/highlights-tokens) package, where we get a list of allowed class names, generates the token list by analyzing the `highlights` package's generated grammar objects. So I added a build step to `atom-language-diff` that generates a `tokens.json` we could use here. This PR pulls that into a separate module along with the stock `highlights-tokens` stuff so we can have a convenient place to add more in the future (**see note below**).
* That all seemed to work, so I added a couple of tests, and here we are.

**Note:** Unless I'm missing something, I believe the other non-stock languages we support for syntax highlighting (`atom-language-nginx`, `language-rust`, etc..) aren't going to make it past the sanitizing step without first going through a similar process of generating their lists of valid tokens. Those might be candidates for new Issues.

### DIFF
--- a/lib/highlights-tokens.js
+++ b/lib/highlights-tokens.js
@@ -1,0 +1,8 @@
+var stockHighlightsTokens = require('highlights-tokens')
+var diffTokens = require('atom-language-diff')
+
+
+// combine stock highlights tokens list with any additional ones we've included
+
+module.exports = [].concat(stockHighlightsTokens, diffTokens)
+
--- a/lib/highlights-tokens.js
+++ b/lib/highlights-tokens.js
@@ -1,8 +1,6 @@
 var stockHighlightsTokens = require('highlights-tokens')
 var diffTokens = require('atom-language-diff')
 
-
 // combine stock highlights tokens list with any additional ones we've included
 
 module.exports = [].concat(stockHighlightsTokens, diffTokens)
-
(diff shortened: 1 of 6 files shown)


zeke:
```diff
- no diff support
+ diff support!
```

aredridel:
Wow. this looks awesome. Nice careful work there!

ashleygwilliams:
this is awesome @revin !! merging

jonathanmarvens:
Nice! LGTM :+1:!

zeke:
Hey @revin how did you find out about github's support for diff syntax highlighting? I'm trying to find a GitHub page that explains it so I can add a link to the readme. I suspect few people know about this great feature of GFM.

wooorm:
@zeke I think it's in the linguist repo, which if I remember correctly houses all there supported fenced code block flags!

revin:
@zeke Honestly I didn't know about it until I read #64 :-P and then went hunting through the `highlights` package to see how it could be done

PR #96, opened 2016-01-11
Wrap code blocks to match former output

revin:
This fixes the tests reported broken in #90 due to a [change in renderer behavior in `markdown-it` 5.1.0](https://github.com/markdown-it/markdown-it/issues/190) by supplying a custom renderer rule that just wraps the output of the built in `fence` rule with `<pre><code class="...">...</code></pre>`. Discussion of the validity of nested `<pre>` elements aside, the effect is to restore the behavior from `markdown-it` 5.0.x as used here in `marky-markdown`.

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -45,8 +45,7 @@ module.exports = function (html, options) {
     }
   }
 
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
closing per decision in #90

Issue #95, opened 2016-01-11
Restrict emoji substitution to colon-wrapped words


cloakedninjas:
As per npm/newww#463 emoticons are being rendered into markdown which doesn't match Github's method. Emoji substitution ideally would only replace colon-wrapped items, e.g. [colon]+1[colon] -> :+1: 

Example error: https://www.npmjs.com/package/rumpus

revin:
`markdown-it-emoji` does the colon-delimited emoji replacement, but it also has this notion of "shortcuts" where it does pattern matching to replace regular text-style smilies with unicode versions as well. PR #97 disables the shortcuts, which if I'm understanding this issue correctly, should be all we need to do.

ashleygwilliams:
closed by #97

ashleygwilliams:
@cloakedninjas thanks so much for this issue. this will be landing in `7.0.0` by the end of the week.

ashleygwilliams:
actually, i'mma make this `6.0.2` since it is fixing a bug.

PR #97, opened 2016-01-11
Disable smiley shortcut handling in markdown-it-emoji

revin:
Here's a tiny PR that disables the "shortcut" handling in the `markdown-it-emoji` plugin we use for emoji support; fixes the issues pointed out as per #95.

This isn't directly compatible with #96 due to the rearranging I did there, but it's a pretty trivial modification to merge between the two; I'm happy to update whatever as needed.

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -45,7 +45,7 @@ module.exports = function (html, options) {
     }
   }
 
-  var parser = MD(mdOptions).use(lazyHeaders).use(emoji)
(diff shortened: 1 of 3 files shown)


ashleygwilliams:
A++ u killing it @revin thanks SO SO SO much for your help

ashleygwilliams:
oh hey, could you write a test tho? if not lemme know and i can do it

revin:
Good call; sorry, got ahead of myself. PR updated.

Issue #71, opened 2015-11-25
make fixes based on code climate suggestions


ashleygwilliams:
https://codeclimate.com/github/npm/marky-markdown/issues/categories/bugrisk

zeke:
I'm not sure about [the suggestion](https://codeclimate.com/github/npm/marky-markdown/bin/marky-markdown.js) to use `throw` instead of `process.exit()`. Doing so will result in a bunch error output to the screen, when all we really want is usage:

```
❯ node bin/marky-markdown.js
/Users/z/zeke/marky-markdown/bin/marky-markdown.js:7
  throw new Error('Usage:\n\nmarky-markdown some.md > some.html')
  ^

Error: Usage:

marky-markdown some.md > some.html
  at Object.<anonymous> (/Users/z/zeke/marky-markdown/bin/marky-markdown.js:7:9)
  at Module._compile (module.js:398:26)
  at Object.Module._extensions..js (module.js:405:10)
  at Module.load (module.js:344:32)
  at Function.Module._load (module.js:301:12)
  at Function.Module.runMain (module.js:430:10)
  at startup (node.js:141:18)
  at node.js:980:3
```

https://twitter.com/zeke/status/684425698321158144

ashleygwilliams:
closing. i don't think these checks are worthwhile.

PR #98, opened 2016-01-12
Move re-scan function into separate function to avoid linter warning

revin:
For #71, since the linter is complaining that we shouldn't create a function in a loop.

My initial reaction is to just replace the anonymous function on line 17 with a reference to `scan`, but since technically the `clean()` function passes the time lived as a parameter to the callback, I decided it was (maybe?) clearer to make sure it's always called with an empty parameter list.

### DIFF
--- a/lib/cleanup.js
+++ b/lib/cleanup.js
@@ -11,12 +11,15 @@ module.exports = function (grammars) {
     while (stops.length) stops.shift()()
   }
 
+  function runScan () {
+    scan()
+  }
+
   function scan () {
     while (managed < grammars.length - 1) {
       stops.push(
-        clean(grammars[managed], 'repository', 2000, function () {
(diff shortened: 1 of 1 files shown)


ashleygwilliams:
mind weighing in on this one @aredridel ?

revin:
reference for my assertion that the callback receives a parameter: https://github.com/soldair/property-ttl/blob/master/index.js#L41

aredridel:
My personal feeling: burn this linter to the ground, because that check is not useful, I would only ever care about that in a performance critical section (and we're talking 'hot game loop' levels of performance not 'eke out a tens of ms from web requests'. I'd disable that lint check because it serves no purpose.

ashleygwilliams:
closing as per @aredridel's comments

revin:
Totally agree, @aredridel.

PR #100, opened 2016-01-12
Update tests to support markdown-it 5.1.0

revin:
As of `markdown-it` 5.1.0, we no longer have a `<pre><code>...</code></pre>` wrapper on fenced code blocks; this PR updates the tests that fail as a result of that change, plus removes a couple that were specifically looking for `class="highlight"` on said `<code>` elements.

If/when this is merged, the root of the rendered code blocks will be `<pre class="editor">`, and lots of dependents' CSS will need updated.

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -45,8 +45,7 @@ module.exports = function (html, options) {
     }
   }
 
-  var parser = MD(mdOptions).use(lazyHeaders).use(emoji, {shortcuts: {}})
-  return parser.render(html)
+  return makeParser(mdOptions).render(html)
 }
 
 var mappings = {
@@ -74,4 +73,21 @@ function scopeNameFromLang (highlighter, lang) {
   // mappings[lang] = name
 
   return name
+}
+
+// create a parser instance, add plugins, update rules, etc...
+function makeParser (mdOptions) {
+  var parser = MD(mdOptions).use(lazyHeaders).use(emoji, {shortcuts: {}})
+
+  // monkey patch the 'fence' parsing rule to restore markdown-it's pre-5.1 behavior
+  // (see https://github.com/markdown-it/markdown-it/issues/190)
+  var stockFenceRule = parser.renderer.rules.fence
+  parser.renderer.rules.fence = function (tokens, idx, options, env, slf) {
+    // call the original rule first rather than inside the 'return' statement
+    // because we need the 'class' attribute processing it does
+    var output = stockFenceRule(tokens, idx, options, env, slf).trim()
+    return '<div' + slf.renderAttrs(tokens[idx]) + '>' + output + '</div>\n'
+  }
+
+  return parser
 }
--- a/lib/render.js
+++ b/lib/render.js
@@ -4,6 +4,7 @@ var MD = require('markdown-it')
 var lazyHeaders = require('markdown-it-lazy-headers')
 var cleanup = require('./cleanup')
 var emoji = require('markdown-it-emoji')
+var codeWrap = require('./code-wrap')
 
 var highlighter = new Highlights()
 
@@ -45,7 +46,11 @@ module.exports = function (html, options) {
     }
   }
 
-  return makeParser(mdOptions).render(html)
+  var parser = MD(mdOptions)
+    .use(lazyHeaders)
+    .use(emoji, {shortcuts: {}})
+    .use(codeWrap)
+  return parser.render(html)
 }
 
 var mappings = {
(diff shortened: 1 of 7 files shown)


ashleygwilliams:
hrm @revin this looks like it has conflicts? not sure what's up

ashleygwilliams:
ooh i bet it is because i publish `6.0.2`

revin:
Oh yeah maybe that's it. I pulled the upstream branch and made the change right away, so I wasn't sure what could've happened in the meantime. If it's something more involved, let me know and I can redo it.

ashleygwilliams:
nah nearly certain that's it. haven't touched anything else: https://github.com/revin/marky-markdown/blob/code-block-unwrap/package.json#L3

revin:
Conflict resolved. However, oops, it looks like an auto-formatting lint step changed some whitespace in `lib/render.js` :oops:

zeke:
Because of the changing classnames here, there will likely need to be a companion update to newww's stylesheets.

ashleygwilliams:
@zeke yup, we talked through all of that in #90 :)

ashleygwilliams:
@revin i'm a little confused about the commit history here... i think a rebase might just fix it? but it looks like each commit happens 2x? (once individually, and then once together?)

revin:
@ashleygwilliams you're right, sorry I should've force pushed it. I think I got it this time.

ashleygwilliams:
there we go! ok, this is a big change so i'm gonna pull it down and play with it, but should merge soon :) thank you thank you thank you :space_invader: :dolphin: :sparkling_heart: :sparkles:

revin:
Yep yep! To be honest, I'm a little nervous about losing the `<code>` part of the old formatting. It's a minor nit, but IMHO we're losing a bit of meaning without it.

ashleygwilliams:
yeah i hear ya @revin -- the product roadmap for marky (if we can be said to have one) is to mimic github markdown parsing as closely as possible as that is what our users expect. so this brings us closer to that, for better or worse :/

aredridel:
Looks good to me.

revin:
*sigh*, I'm really really sorry to belabor this, but something just doesn't feel right. We provide `markdown-it` with a `langPrefix: 'highlight '` but then it doesn't appear in the rendered markup at all because it skips the wrapping elements that would have had that class name attached.

I think the issue is that `markdown-it` is intending to treat rendered markup coming back from the highlighter with a leading `<pre` as complete, so it doesn't do any more processing to it. But the `highlights` package is intentionally only rendering the content of whatever code we provide, leaving it up to the calling code to decorate or contain it as necessary. So there's a little bit of a mismatch between the two libraries.

Github wraps their `<pre>` with `<div class='highlight highlight-source-[lang]'>`, so should we follow suit (via a custom rule similar to what #96 had)?

revin:
Oh, and doing the wrapping DIV ourselves would also lead to resurrecting those two deleted tests, plus having to update the sanitizer module to account for the markup changes.

ashleygwilliams:
hey @revin thanks for your thoughtful insight on this.
as stated previously, our only goal is to bring marky in alignment with GH-flavored markdown. to that end i think your suggestion of adding the div wrapper is a good one.

thoughts @aredridel @jefflembeck ?

ashleygwilliams:
hey @revin looks like you updated this but since it was an amend i'm not sure exactly what changed-- can u comment? thanks so much :smile:

revin:
Nothing functionally changed, I just rebased onto master because one of the package.json changes recently created a small merge conflict. I have the div wrapper change ready locally if we need it whenever we make the final decision on that.

ashleygwilliams:
if it moves us toward GH-style i say let's do it. `7.0.0` will have breaking changes for `newww` anyways so they'll be updating code no matter what.

revin:
ok @ashleygwilliams, latest change is pushed and Travising. I'm totally open to suggestion on the best selectors to use for the tests; I switched from requiring a `<code>` element to requiring any old element with `class='highlight'` in most of them, but it's entirely possible there's something semantically better to use.

zeke:
Looks good to me.

lib/render.js
```diff
@@ -75,3 +74,20 @@ function scopeNameFromLang (highlighter, lang) {
...
 }
+
+// create a parser instance, add plugins, update rules, etc...
+function makeParser (mdOptions) {
+  var parser = MD(mdOptions).use(lazyHeaders).use(emoji, {shortcuts: {}})
+
+  // monkey patch the 'fence' parsing rule to restore markdown-it's pre-5.1 behavior
+  // (see https://github.com/markdown-it/markdown-it/issues/190)
+  var stockFenceRule = parser.renderer.rules.fence
+  parser.renderer.rules.fence = function (tokens, idx, options, env, slf) {
```
revin: @ashleygwilliams this pattern here is basically the same thing [markdown-it-expand-tabs](https://github.com/revin/markdown-it-expand-tabs/blob/master/index.js#L12-L16) does. If you think it would be cleaner, I could extract this into a super tiny markdown-it plugin. We could either publish it separately or just include it with marky in a separate module, and use markdown-it's regular `.use()` API to hook it up.

aredridel:
LGTM!

ashleygwilliams:
yar this is my bad, but this is gonna need a rebase. i want to release a 6.0.4 today with some of the new tests and things. and then try to get 7.0.0 out the door for tomorrow. again, my bad :(

revin:
Don't even worry about it; rebased and travis'd.

PR #102, opened 2016-01-12
Fix code example in README.md

latentflip:
Was just missing a couple of commas so it wouldn't run when I copy-pasted.


zeke:
LGTM

ashleygwilliams:
great! i'm looking at redo-ing marky's docs (take them out of the readme and into a webpage of some sort) thanks for this!

Issue #48, opened 2015-05-04
README h1s not stripped for scoped packages


sindresorhus:
https://www.npmjs.com/package/@sindresorhus/df

Not sure why. Maybe because of the short header.

zeke:
Are you saying you expected the `# df` H1 line from your readme to be hidden?

sindresorhus:
Yes

ashleygwilliams:
Adding this to `5.7`.

revin:
Is it because the scope name wrecks the [similarity calculation](https://github.com/npm/marky-markdown/blob/master/lib/packagize.js#L14-L15)? I've got a change locally that also tests the similarity of the H1 to the package name sans scope, but I'm having trouble finding any other scoped packages in the NPM registry to test against. @ashleygwilliams do you know how I can get a list of a few?

ashleygwilliams:
ooh @revin i think you are on to something there. p sure that is :100: it.
here's another example: https://www.npmjs.com/package/@aredridel/redis-mock

ashleygwilliams:
@sindresorhus think we've found the source of the issue. ideally the fix will ship with 7.0.0 and be on npmjs.com shortly after.

revin:
Thanks @ashleygwilliams, that helped. PR is #103, tiny update ftw I guess 😜

ashleygwilliams:
fixed in https://github.com/npm/marky-markdown/pull/103, closing

PR #103, opened 2016-01-12
Ignore package scope when calculating header/name similarity

revin:
I think the issue in #48 is that the similarity checker has no knowledge of scoped package names. If we add a line of code that does a regex match to filter out anything between a leading `@` and the first `/` (is that logic correct? Please double check my assumptions), then the packagizer code correctly applies `class='package-name-redundant'` to the generated H1.

I tested this against [@sindresorhus/df](https://www.npmjs.com/package/@sindresorhus/df) and [@aredridel/redis-mock](https://www.npmjs.com/package/@aredridel/redis-mock).

### DIFF
--- a/lib/packagize.js
+++ b/lib/packagize.js
@@ -12,6 +12,7 @@ var packagize = module.exports = function ($, pkg) {
   var h1 = $('h1').first()
   if (
     similarity(pkg.name, h1.text()) > 0.6 ||
+    similarity(pkg.name.replace(/^@[^\/]+\//, ''), h1.text()) > 0.6 || // filter out scope name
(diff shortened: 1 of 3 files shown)


ashleygwilliams:
this is great, but i'm confused as to why your PRs aren't getting run on travis. investigating.

revin:
whoops, if there's something I need to do differently, please let me know 😕

ashleygwilliams:
could you try pushing an empty commit to this? want to see if what i just did fixed it

ashleygwilliams:
THERE WE GO. feel free to rebase that out now :)

revin:
WOO! :+1:

ashleygwilliams:
looks great, merging!

PR #106, opened 2016-01-13
Update to lodash 4.0, handle its breaking changes

revin:
Apologies if I'm pushing ahead too quickly on this, but greenkeeper reported in #105 that switching to lodash 4 breaks the build; turns out all we need to do is call `_.pickBy()` rather than `_.pick()` when we're resolving languages in the code highlighting stuff.

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -62,7 +62,7 @@ function scopeNameFromLang (highlighter, lang) {
 
   if (mappings[lang]) return mappings[lang]
 
-  var grammar = _.pick(highlighter.registry.grammarsByScopeName, function (val, key) {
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
same travis issue here. push an empty commit or something else and rebase to trigger it.

revin:
Hmm, I did, and the Travis info appeared here, but then it went away when I rebased. Does it break if I do it too quickly?

revin:
Maybe it had to do with the way I rebased. Straight up deleting the empty commit seemed to stop the CI build, but squashing it worked OK...?

ashleygwilliams:
i think someone messed with marky's travis settings :/ i found them changed when i went in just now.

ashleygwilliams:
ok, this looks great, merging.

PR #107, opened 2016-01-13
Update README's dependencies list to reflect current project state

revin:
The dependencies listed in the README aren't all up to date, so here's a commit to fix that.


ashleygwilliams:
A++

PR #110, opened 2016-01-13
Tighten up the YouTube filter regex

### LINKED ISSUE #108: all iframes are allowed
https://github.com/npm/marky-markdown/blob/b6e752bf799d427b19a792515710321b81299d12/lib/sanitize.js#L57-L59

The comment says that it allows only iframes from youtube. In fact, it allows iframes from anywhere, as long as the URL contains `//youtube.com`.

revin:
Prior to this commit, any URL containing "//youtube.com" would make it through our filter, but the intent was to only allow actual YouTube URLs. So now we're more explicit about requiring that "//youtube.com" either starts the string, or is preceded by "http:" or "https:". We've also got some test updates to make sure URLs previously wrongly allowed get filtered out, plus that a full URL where the "//" doesn't appear at the very beginning still works.

Fixes #108.

### DIFF
--- a/lib/sanitize.js
+++ b/lib/sanitize.js
@@ -59,6 +59,6 @@ sanitizer.config = {
   exclusiveFilter: function (frame) {
     // Allow YouTube iframes
     if (frame.tag !== 'iframe') return false
-    return !String(frame.attribs.src).match(/\/\/(www\.)?youtube\.com/)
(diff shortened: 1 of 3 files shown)


ashleygwilliams:
looks great! def an improvement on what we had previously. :dart:  merging.

PR #114, opened 2016-01-15
Overhaul devDependencies' readme parsing test strategy

### LINKED ISSUE #91: move devdeps that exist for testing into static fixtures
i want to use greenkeeper and this gets in the way
related #86

revin:
A bunch of our `devDependencies` are only included so we can use their README data as test fixtures. This commit adds their READMEs as static files we load separately during `npm test` and gets rid of their entries in `package.json`. As far as I can tell, we only need `mocha`, `standard`, and `glob` for actual development.

The overall testing strategy is similar to what we already have, except instead of loading all of the markdown strings in `fixtures.packageNames` we split them into `fixtures.dependencies` (which may be a poor name, since it *does* also include the `devDependencies` as well) and `fixtures.examples`. This allows us to differentiate them during test runs because we don't have to try to `require()` the static fixtures' packages; we can just mock them as `{name: name}`.

Unless there's more to this than I could glean just by reading the issue itself, this fixes #91.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,20 +52,9 @@
     "similarity": "^1.0.1"
   },
   "devDependencies": {
-    "async": "^0.9.0",
-    "benchmark": "^1.0.0",
-    "cicada": "~1.1.1",
-    "express": "^4.10.7",
     "glob": "^4.3.5",
-    "grunt-angular-templates": "^0.5.7",
-    "johnny-five": "^0.8.37",
-    "maintenance-modules": "1.0.2",
-    "memoize": "~0.1.1",
-    "mkhere": "~1.0.9",
     "mocha": "^2.0.1",
-    "payform": "1.0.1",
-    "standard": "^3.7.0",
-    "wzrd": "^1.1.1"
+    "standard": "^3.7.0"
   },
   "bin": "./bin/marky-markdown.js",
   "contributors": [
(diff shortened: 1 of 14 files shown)


ashleygwilliams:
YOU ARE MY HERO ZOMG I HAVE BEEN PUTTING THIS OFF FOR DAYS

ashleygwilliams:
:muscle: :sparkles:

revin:
🎉 haha me too 😁

zeke:
fabulous.

revin:
hmm, I was gonna split this commit into two: one to add the static fixtures, and one to update the code, but then I realized having the new directory `test/fixtures/readmes` ends up breaking the [process of loading the hand-written fixtures](https://github.com/npm/marky-markdown/blob/master/test/fixtures.js#L9). Does anyone have a problem with the build being broken for one commit as long as things are still working after the PR is merged?

(@ashleygwilliams feel free to answer "LOL@policy" again)

revin:
oops looks like we need `glob` too

ashleygwilliams:
i dont mind you combining them unless you've already done the separation. i can manage the release either way. ideally, i like every PR to be a complete "makes dat travis boy happy" feature. but also yes (LOL@policy)

revin:
phew ok Travis is green, so unless anyone prefers a different strategy for this, I think we're good to go.

ashleygwilliams:
i'll need a second to review but honestly if it gets @greenkeeper of my back lol ..

ashleygwilliams:
this looks great and i'm super excited about it. thanks so much @revin!

PR #116, opened 2016-01-15
Remove frontmatter processing code

### LINKED ISSUE #43: Remove unused frontmatter-to-meta-tag code
We're not actually using the generated meta tags, and it's causing problems: https://github.com/npm/newww/issues/850

revin:
If nobody is using the meta tags we create from README frontmatter, we may as well just remove that code altogether since it has the potential to generate unexpected results.

For test coverage, all I did was remove the lone test that was looking for meta tags. If you'd rather me test that we're specifically doing nothing in response to leading comments in markdown files, let me know.

Fixes #43.

(note: this is a speculative PR based on a shallow reading of the issue, which has been around this project far longer than I have!)


zeke:
I'm all for it!

ashleygwilliams:
removing code is awesome. merging! :dog2: :money_with_wings: :sparkles:

Issue #122, opened 2016-01-21
break test file up into different files by category


ashleygwilliams:
i really dislike huge files. would like to see the tests broken down by category.

revin:
:+1: I was just thinking about doing this yesterday in fact

ashleygwilliams:
hehe yeah, super long files are a huge pet peeve of mine :)

ashleygwilliams:
solved by #123

PR #123, opened 2016-01-22
Split unit tests into a series of smaller files

revin:
Per #122, this splits the tests into a bunch of smaller files.

I'm enough of a n00b at Node conventions that I don't know if it's generally accepted to create modules lacking a `module.exports`, but doing so keeps the code a little simpler. The test groups themselves were pretty well defined already, so this is mostly just a copy/paste job, plus top-of-file `require` stuff.


revin:
This doesn't include the test changes from #100, so if that gets merged to master, I'll have to update the corresponding tests here as well.

ashleygwilliams:
@revin would you prefer i merge this now or later? s'up to you :)

revin:
There's a little bit of effort either way, so I'd vote for sooner rather than later, if it's all the same to you. Go for it.

ashleygwilliams:
ok! i'll merge this now. thanks :)

Issue #126, opened 2016-01-22
Markdown Rendering Issues on npmjs.com


jhuckaby:
I have discovered a number of Markdown rendering bugs on npmjs.com, which I have compiled into a package with examples:

https://www.npmjs.com/package/npm-markdown-bugs

Compare this to the same exact Markdown syntax on GitHub, and you can see the issues:

https://github.com/jhuckaby/npm-markdown-bugs

I figured this would be better than entering 5 separate issues. But if you would prefer that, please let me know.

Thanks!

ashleygwilliams:
thanks so much for filing :) and extra thanks for the great test module!

revin:
"Incorrect Code Indents" and "Tabs Are Rendered Extremely Wide" might be candidates for marky fixes; essentially the problem there is that browsers are rendering `\t` characters with traditional eight-character width. Github is replacing them with four spaces. We should probably look around to determine under which circumstances they make this substitution (it could just be that tabs in leading whitespace get replaced) and follow suit.

"Incorrect Nested List Indents", "No Space After Lists", and "Tables Centered By Default" look to me like newww CSS issues, not markdown-to-HTML errors.

jhuckaby:
Thanks @revin -- I'll resubmit "Incorrect Nested List Indents", "No Space After Lists", and "Tables Centered By Default" as CSS rendering issues in the https://github.com/npm/newww repo.

ashleygwilliams:
thank you so much @jhuckaby ! i was about to do this. so awesome that you already have :)

jhuckaby:
My pleasure @ashleygwilliams :+1:

revin:
@jhuckaby ditto on the thanks, I really like the idea of having a markdown torture test module specifically there to look for rendering bugs. (And sorry to keep bouncing you around between issue trackers!)

jhuckaby:
Haha, no problem @revin -- happy to help.  I'll continue to add test cases to my torture test module as I find them 😉

revin:
The marky-specific parts of this were fixed in #127; closing

PR #127, opened 2016-01-22
Convert leading tabs to spaces in fenced code blocks

revin:
In #126 we have, among other things, a couple of rendering issues caused by the fact that [markdown-it](https://www.npmjs.com/package/markdown-it) (probably correctly) doesn't do anything with tab characters in fenced code blocks; *and* browsers render them with the traditional eight character width, and sometimes visually collapse sequential tabs into one. Neither of these behaviors matches what GitHub does, which appears to be four-space substitution per leading tab.

I couldn't find anything in the markdown-it plugin ecosystem that already covered this, so I built a markdown-it plugin called [markdown-it-expand-tabs](https://www.npmjs.com/package/markdown-it-expand-tabs) that we can use to help fix these rendering glitches. It's on NPM now, so this PR pulls it in and updates the `basic.md` test fixture to include some tab-indented code we can translate and inspect.

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -5,6 +5,7 @@ var lazyHeaders = require('markdown-it-lazy-headers')
 var cleanup = require('./cleanup')
 var emoji = require('markdown-it-emoji')
 var codeWrap = require('./code-wrap')
+var expandTabs = require('markdown-it-expand-tabs')
 
 var highlighter = new Highlights()
 
@@ -50,6 +51,7 @@ module.exports = function (html, options) {
     .use(lazyHeaders)
     .use(emoji, {shortcuts: {}})
     .use(codeWrap)
+    .use(expandTabs, {tabWidth: 4})
   return parser.render(html)
 }
 
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "markdown-it": "^5.1.0",
     "markdown-it-lazy-headers": "^0.1.3",
     "markdown-it-emoji": "^1.1.0",
+    "markdown-it-expand-tabs": "^1.0.7",
     "property-ttl": "^1.0.0",
     "sanitize-html": "^1.6.1",
     "similarity": "^1.0.1"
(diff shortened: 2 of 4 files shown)


revin:
Also, I'm passing the plugin a hard-coded tab width of 4 spaces because that's how Github renders. However, it wouldn't be particularly difficult to also allow this to be overridden in marky's `options` parameter if we wanted to do that (and if so, please suggest a good name 😀).

revin:
Looks like the plugin might need some work; it's bombing out of Travis on older Node versions.

revin:
All right, Travis went green; I'll squash and re-push

revin:
[npm/newww#1678](https://github.com/npm/newww/issues/1678) suggests using a package's .editorconfig for tab width. That might be a cool addition to this, or maybe something for soon after.

lib/render.js
```diff
@@ -45,7 +46,7 @@ module.exports = function (html, options) {
     }
   }
 
-  var parser = MD(mdOptions).use(lazyHeaders).use(emoji, {shortcuts: {}})
+  var parser = MD(mdOptions).use(lazyHeaders).use(emoji, {shortcuts: {}}).use(expandTabs, {tabWidth: 4})
```
ashleygwilliams: am i about to argue about tab spacing?
i'm about to argue about tab spacing.

thoughts on 2 vs 4? i'm a 2.

revin: 2 seems to be more Node-y, but from what I'd seen on github, they go with 4 on their markdown rendering, so I chose 4 here. I'm happy to go with whatever you like. "GitHub parity" doesn't necessarily have to mean 100% identical whitespace 😀

ashleygwilliams: hahahha no no, you (and GH parity) win. 4 it is!

ashleygwilliams:
is this still WIP @revin? i fear i've lost track of it. also-- this is a breaking change ya? tentatively setting at 7.0.0

revin:
@ashleygwilliams no longer WIP, this should be good to go for now. I think this is *technically* a breaking change, but the main thing that would probably break is people's CSS for tab widths (unless someone somewhere is attempting to parse the rendered code blocks and is sensitive to the whitespace format), so it's pretty benign. 7.0 sounds like a reasonable milestone to me though.

PR #132, opened 2016-01-28
Add more robust CLI that provides rendering options

knksmith57:
Exposes parse options to the CLI utility giving users a bit more control about how HTML is rendered.

Made use of @bcoe's [`yargs`](https://www.npmjs.com/package/yargs) package for the heavy lifting (figured this would make y'all more likely to accept the PR :wink:). This does add another dependency to the package, but IMHO it's worth the extra weight to make the CLI much more usable.

Cheers!


zeke:
niiiice

zeke:
If you want to add tests to this, I have a few ideas: You could use [sinon](http://npm.im/sinon) to spy on the marky function and ensure the right options are being passed, or you could use [nixt](http://npm.im/nixt) to actually run CLI commands and verify their `stdout` output.

revin:
Oooh I like this. At the very least it'll make my own testing easier 😉

bcoe:
yarrrrgs \o/ If you decide to add some tests for the CLI behavior, here's the approach we use in the yargs codebase itself:

https://github.com/bcoe/yargs/blob/master/test/integration.js

we use this along with: 

https://github.com/bcoe/nyc

Which gives good code coverage for subprocesses.

ashleygwilliams:
hey @knksmith57 this is super awesome! would love to get some tests for this before merging tho, as well as some docs in the readme. lemme know if you want any help! thanks for doing this :)

knksmith57:
Hey All, 

Thanks for the feedback! I'll jam on some test cases and docs and update the PR once I've got something worth looking at.

ashleygwilliams:
hey @knksmith57 just wanted to say that we are all still really excited about this. let us know if you want any help :space_invader:

knksmith57:
Hey @ashleygwilliams! Sorry for the radio silence :disappointed: 

I pulled in all of the great work y'all have been doing since I opened the PR and then refactored a bit to make the CLI portion easier to test. I've never tested integration between a CLI wrapper & main module before so definitely looking for feedback on this (maybe it's totally wrong!)

Maybe @bcoe could chime in when he has a chance to see if this is a decent / terrible pattern for testing CLIs that use yargs?

Thanks team! Looking forward to jamming on this and getting it shipped :smiley:

Issue #128, opened 2016-01-23
Emoji in headers is stripped from anchors


chrisdickinson:
Hi! I've been poking at writing some documentation, and in the course of playing around with another tool I ran into a discrepancy between how GitHub generates heading slugs vs. how npm renders them, in particular when emoji appears in the header.

I've created a [test repo](https://github.com/chrisdickinson/emoji-slug-example) and [published it](http://npm.im/emoji-slug-example). It appears that npm strips emoji from the anchor entirely, while GitHub includes the emoji's GH nickname in the anchor.

This isn't blocking anything for me; this issue is by way of cataloguing the differences between GH's rendering and ours — apologies if this is a known thing. If so, please don't hesitate to close this out!

Thanks, all!

ashleygwilliams:
thanks for reporting @chrisdickinson !

revin:
Yes, thanks @chrisdickinson!

OK @ashleygwilliams I filed a PR on github-slugger since that's what we use to generate the slugs and it made sense to me to have the emoji code live there. It's [github-slugger#2](https://github.com/Flet/github-slugger/pull/2).

I'm somewhat amused that much of the time, the emojis will start as `:shortcode:`-style strings in the markdown, then get converted to HTML via markdown-it, then get converted *back* to shortcodes with the stuff I PR'd for github-slugger. But given that the alternative was to be more invasive and hackier in markdown-it, it seemed to me like the cleanest way to do it. As always, I'm totally open to critique/disagreement/reassessment.

revin:
All right, they declined the PR and recommended we make the change on our side, so I submitted PR #133 here to do just that.

ashleygwilliams:
iirc github slugger was made for this repo. if they wont take it, lets fork. (will review when i am no longer on planebut that is my initial thought)

zeke:
cc @Flet

wooorm:
Hey all! I’m the one who “declined” / advised against the pull, not @flet, it’s his project so if he wants he should definitely merge.

However, the reason for advising against is because the proposed solution actually breaks other things on npm. Sorry, if that didn’t come across.

Here’s some more in depth description of what I tried to say over at flet/github-slugger#2.

There’s a different slugging happening between the following two headers on GitHub:

```md
# 😄 - an emoji

# :smile: - a gemoji
```

[...]

PR #133, opened 2016-01-29
Convert emoji to shortcode names in generated heading IDs

revin:
This PR converts emoji characters in heading IDs back to shortcode names prior to running them through the `github-slugger` module, in an effort to make marky's generated heading IDs more closely match the ones GitHub generates. The included tests are based on @chrisdickinson's test cases in #128, thanks for those!


test/headings.js
```diff
@@ -39,31 +39,31 @@ describe('headings', function () {
 
   it('allows a dash in generated DOM ids just like GitHub', function () {
     assert(~fixtures.github.indexOf('### heading with a - dash'))
-    $ = marky(fixtures.github)
+    $ = marky(fixtures.github, {prefixHeadingIds: false})
```
revin: I'm curious as to why these tests worked before; since they were lacking `{prefixHeadingIds: false}`, shouldn't they have been generating the `user-content-` prefix in the IDs? If not, then I might have something wrong in my logic and will need to amend my changes.

test/markdown.js
```diff
@@ -75,12 +75,6 @@ describe('markdown processing and syntax highlighting', function () {
     assert($("a[href='https://gist.github.com/sindresorhus/8435329']").length)
   })
 
-  it('replaces markdown syntax for emoji with unicode for the emoji', function () {
```
revin: This test isn't deleted, just moved into the new `emoji.js` test suite.

lib/headings.js
```diff
@@ -1,37 +1,68 @@
-var fmt = require('util').format
+var cheerio = require('cheerio')
```
revin: This `headings.js` file is now a plugin for `markdown-it` rather than a standalone Node module for the rest of marky to consume (the only substantial effect of this change is to move the one line where the module is used from `index.js` to `render.js`).

@ashleygwilliams do you think it's worth splitting into a completely separate package and publishing it on NPM? It's pretty specific to marky, since it's basically "a markdown-it plugin for generating GitHub-like headings for applications that *also* use the markdown-it-emoji plugin"

revin:
I'm pretty happy that this code even handles the pathological case of a heading containing unicode emoji, text gemoji, *and* embedded HTML/markdown all at the same time. :+1:

(If anyone knows of one of those out in the wild, please let me know)

lib/token-util.js
```diff
@@ -0,0 +1,36 @@
...
+// Look for tokens with type === 'emoji', which are :shortcode: style emoji
+// characters that have been replaced by the markdown-it-emoji plugin. Return the
+// original tokens, unless they were converted gemoji strings; then return a copy so
+// we haven't clobbered the original when it comes time to render HTML
+function unconvertEmoji (Token) {
+  return function (token) {
+    var tok = token
+    if (token.type === 'emoji') {
+      tok = assign(new Token(), token)
+      tok.meta = token.content
```
revin: I just realized I don't need this line anymore now that I'm using lodash.assign to make copies; was an artifact of an earlier attempt. I'll remove it when I get a chance.

ashleygwilliams:
i'm happy keeping it in marky for now, i think
also this is rad thank you
waiting on 7.0.0 for this even though it is a bug fix, i like to be safe :)

ashleygwilliams:
oh also, this needs a rebase :)

revin:
ok, done.

Speaking of markdown-it-pluginizing stuff, most (all?) of our other processing steps that we currently do on the generated cheerio DOM object could also be converted to markdown-it plugins. Intuitively it would be faster and/or more efficient to do it since we'd be operating all at once during the markdown parsing phase (though I don't have any data to determine that one way or another for real), but the downside of a markdown-it plugin is that it doesn't process embedded HTML, so any operations we do there will skip it.

For example, the thing where we force the protocol to 'https://' for gravatar images: as a markdown-it plugin, it would skip any inline HTML `` elements because those aren't recognized as actual markdown tokens internally.

My assumption is that this is enough of a problem that we should just keep things as is. However, if we *did* move everything into markdown-it, then some future version of marky could just return a string of HTML rather than a cheerio DOM object, saving that whole step and its associated resource usage. If most of marky's users are just doing `marky(...).html()` anyway, that might be something to consider. We could always create, say, `marky(...).dom()` for people that wanted the DOM object.

*Anyway*, that was a thought that occurred to me while I was doing this, so I figured I may as well say something in case anyone thought that was interesting.

lib/token-util.js
```diff
@@ -0,0 +1,46 @@
...
+  if (idx < 0) {
+    token.attrPush([attr, value])
+  } else {
+    if (operation === 'append') {
+      token.attrs[idx][1] += ' ' + value
+    } else {
+      token.attrs[idx][1] = value
+    }
+  }
+}
```
revin: Once #100 is merged, this function can disappear, and our [headings plugin](https://github.com/revin/marky-markdown/blob/emoji-in-heading-anchors/lib/headings.js#L40-L43) can switch to using markdown-it 5.1's new [`attrJoin`/`attrSet` methods](https://github.com/markdown-it/markdown-it/commit/18f1531024604c090c16233ee44c572be3115ee0). @ashleygwilliams Should that happen in 7.0.1, or should I rebase this PR onto the other one so it all happens in 7.0.0?

ashleygwilliams:
hey @revin i think your point above is important enough to make its own issue-- wanna file? i think it's def something we should look at re perf

lib/token-util.js
ashleygwilliams: rebase onto the other, we're gonna land em together i think (unless you or anyone else has objections)

PR #134, opened 2016-01-30
Remove final references to frontmatter

revin:
We still had `html-frontmatter` listed as a dependency in the README, and a `frontmatter` keyword in package.json, but those are out of date now, no? This commit cleans them up.


ashleygwilliams:
LGTM!

PR #138, opened 2016-02-03
Increase test coverage

revin:
I decided to take a look at our test coverage to see if there were any substantial gaps, so I ran our tests through istanbul and found just a small handful of lines that weren't covered. The changes in this PR cover most of it.

**Please note that I didn't actually add the code coverage machinery.** It's certainly something we could do if everybody wanted to, but I was just after the data so I ran it as a one-off by doing `npm install istanbul` followed by `./node_modules/istanbul/lib/cli.js cover node_modules/mocha/bin/_mocha -- -R spec` and opening `./coverage/lcov-report/index.html` in a browser.

The changes:
- Probably the most important one is where our `packagize` module would error out if the package provided lacked a name (because it would try to do `pkg.name.replace(...)` on an undefined value). I added a quick `if()` guard
- We weren't fully testing that the `cdn` module would bail out if there wasn't sufficient package data supplied, so I added a few tests for that (this is what revealed the issue in `packagize`)
- Added a few lines to the `github.md` test fixture to make sure it properly handles `` elements with a blank/missing `src` attribute and `<a>` elements with a blank/missing `href`
- Ensure running marky with `{debug: true}` produces the same output as normal execution

After these updates, the only code not covered by tests is in the `cleanup` module:

screen shot 2016-02-03 at feb 3 2 13 33 pm

screen shot 2016-02-03 at feb 3 2 12 12 pm

screen shot 2016-02-03 at feb 3 11 38 06 am

That code never gets exercised in my normal development/testing cycle, but I suppose we could come up with a way to make it happen if you thought it was important enough. Everybody's got their own opinions about how important code coverage is, so I'll just defer to the majority.

### DIFF
--- a/lib/packagize.js
+++ b/lib/packagize.js
@@ -7,6 +7,7 @@ var markdown = require('markdown-it')({html: true})
 
 var packagize = module.exports = function ($, pkg) {
   if (!pkg) return $
+  if (!pkg.name) return $
 
   // mark first h1 if it closely matches pkg name
   var h1 = $('h1').first()
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
   },
   "devDependencies": {
     "glob": "^6.0.4",
+    "intercept-stdout": "^0.1.2",
     "mocha": "^2.0.1",
     "standard": "^5.4.1"
   },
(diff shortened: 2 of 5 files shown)


ashleygwilliams:
zomg this is really awesome @revin.

PR #144, opened 2016-02-08
update to standard@6.0.4

Flet:
Closes #139 
Closes #140 
Closes #141
Closes #142 
Closes #143 

Hello! This PR updates `marky-markdown` to the new major version of `standard`. 

The two notable updates for this repo are:
- `standard-format` must not be installed and used separately. Updated the test script to run it along with `standard`
- The file `fixtures.js` was updated to abide the new rule: 
  - `Use path.join() or path.resolve() instead of + to create paths. (no-path-concat)`

Cheers!

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "npm's markdown parser",
   "main": "index.js",
   "scripts": {
-    "test": "standard --format && mocha --timeout 5000"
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
woo this is awesome! thanks so much!

PR #151, opened 2016-02-10
Github-style host/domain linkification

revin:
This change moves us toward parity with Github's linkification of raw domain/hostnames by disabling the `fuzzyLink` option in markdown-it's linkification code, which is the code responsible for creating links from domains/hostnames without a protocol specifier (i.e., *example.com*). It will still create normal links from fully-qualified URLs (*http://example.com/*) and `mailto:` from email addresses (*marky@example.com*). Since GH linkifies hostnames starting with `www.` (www.example.com), we need to add our own schema/rule covering this case.

**Note:** This isn't done; I'll remove the WIP label when it's ready for review.

### DIFF
--- a/lib/linkify.js
+++ b/lib/linkify.js
@@ -1,0 +1,84 @@
+// github-style linkification for linkify-it
+//
+// Pass in a markdown-it parser instance and this will return a version fixed
+// up for the kind of linkification we're after.
+//
+// As far as I can determine, the basic algorithm is to linkify anything
+// starting with 'www.' and ending with a) a whitespace character, or b) a dot
+// followed by a whitespace character (in other words, include internal dots,
+// but exclude a trailing dot). There are plenty of cases where this results in
+// invalid links, but that's how they do it, so that's how we'll do it.
+//
+
+function isInBrackets (text) {
+  return text.charAt(0) === '[' && text.charAt(text.length - 1) === ']'
+}
+
+function isInParens (text) {
+  return text.charAt(0) === '(' && text.charAt(text.length - 1) === ')'
+}
+
+module.exports = function (parser) {
+  if (parser) {
+    var linkify = parser.linkify
+
+    linkify.set({fuzzyLink: false}) // turn off auto-linking normal hostnames
+    linkify.add('//', null) // turn off protocol-relative links
+
+    // linkify everything hostnamey starting with 'www.', optionally in [] or ()
+    var scheme = {
+      validate: function (text, pos, self) {
+        if (!self.re.githubLinkify) {
+          self.re.githubLinkify = new RegExp('^' + self.re.src_host_port_strict + self.re.src_path, 'i')
+        }
+
+        // chop off any trailing ']' or ')', if necessary
+        var candidate = text.slice(pos)
+        if (isInBrackets(text) || isInParens(text)) {
+          candidate = candidate.slice(0, candidate.length - 1)
+        }
+
+        if (self.re.githubLinkify.test(candidate)) {
+          return candidate.match(self.re.githubLinkify)[0].length
+        } else if (candidate.charAt(0) === ' ') {
+          // if we have a zero-length match, i.e., if the original string was
+          // literally 'www.' all by itself, we return -1 because linkify-it does
+          // a boolean test on the result of the validator function (so we need a
+          // non-zero result here to make it through that test) and adds the
+          // returned value to an index it's using to point to the end of the
+          // matched substring. In this case, that means we'll move backwards one
+          // character and only link 'www' rather than the full 'www.' that was
+          // originally matched
+          return -1
+        }
+        return 0
+      },
+      normalize: function (match) {
+        // since we have 3 different "schemes" in place, our matched URLs might
+        // have a leading bracket or parenthesis; trim it off if necessary
+        if (match.schema.charAt(0) === 'w') {
+          match.url = 'http://' + match.url
+        } else {
+          match.url = 'http://' + match.url.slice(1)
+        }
+        match.schema = ''
+      }
+    }
+
+    parser.linkify.add('www.', scheme)
+    parser.linkify.add('[www.', scheme)
+    parser.linkify.add('(www.', scheme)
+
+    // linkify-it's default behavior is to create matches against the
+    // registered schema prefixes when the character immediately preceding a
+    // potential match is non-alphanumeric; however, we also need to skip
+    // linkification when said character is punctuation. So here we rebuild the
+    // internal regular expressions the LinkifyIt instance will use.
+    var regexes = linkify.re
+    regexes.schema_test = RegExp(regexes.schema_test.source.replace(regexes.src_ZPCc, regexes.src_ZCc), 'i')
+    regexes.schema_search = RegExp(regexes.schema_search.source.replace(regexes.src_ZPCc, regexes.src_ZCc), 'ig')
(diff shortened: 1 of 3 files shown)


revin:
Judging by how GH just rendered my text above, the full www.example.com should in fact be linked. I wonder if something like test.example.com works...? :eyes:

revin:
Did a bit of research. The `www.`-linkage part of this task isn't particularly straightforward; there's no official way to modify this part of linkify-it's behavior because it's baked in; it's not part of the installable "schemas" approach it uses for text tokens that have a defined prefix (like a protocol identifier, or "@" for a twitter mention, etc). Unless y'all want to wait for the maintainer to accept/implement this as an option, the next best thing to do is to create a minimal patch for it ourselves and install it at runtime. So that's my next task on this.

revin:
Started an issue at [markdown-it/linkify-it#29](https://github.com/markdown-it/linkify-it/issues/29) to see what the maintainer thinks of making it an official option.

revin:
All right, he wasn't into it, so we're on our own.

The good news is, his response made me realize that rather than hacking on the `fuzzyLink` behavior code, we can treat `www.` as a schema of sorts. As far as I can tell, in addition to linkifying fully qualified URLs, github seems to only ever linkify strings starting with `www.`, regardless of whether they're in fact valid hostnames.

Care to review, @ashleygwilliams?

lib/gh-linkify.js
```diff
@@ -0,0 +1,57 @@
...
+    // linkify everything hostnamey starting with 'www.'
+    parser.linkify.add('www.', {
+      validate: function (text, pos, self) {
+        var tail = text.slice(pos)
+        var regex = /^\S*/
+        var match = tail.match(regex)[0]
+
+        // there's probably a nice regex way of excluding a trailing dot from
+        // these matches, but it's simple enough to just trim them off here
+        if (match.slice(-1) === '.') match = match.slice(0, -1)
```
revin: If anyone knows the regex fu that will allow internal dots in a linked host, but exclude a dot if it's the last thing that appears before any whitespace, we can fold that into the regex on line 23 and get rid of this line. I tried a few things but decided this was clearer (for me, at least).

ashleygwilliams:
LGTM. what are you thinking here? is this a bug fix? or breaking? we have a `7.0.1` and an `8.0.0` coming up so either is fine.

lib/linkify.js
```diff
@@ -0,0 +1,55 @@
...
+        match.url = 'http://' + match.url
+      }
+    })
+
+    // linkify-it's default behavior is to create matches against the
+    // registered schema prefixes when the character immediately preceding a
+    // potential match is non-alphanumeric; however, we also need to skip
+    // linkification when said character is punctuation. So here we rebuild the
+    // internal regular expressions the LinkifyIt instance will use.
+    var re = linkify.re
```
ashleygwilliams: only comment is the `re` is really short and maybe it's more readable to say `regex` or something a bit more elaborate. don't really have strong feels tho. the comment is super helpful.

revin:
I personally feel like this is a bug fix IMHO, since it changes the current wrong behavior people were reporting that our link generation is too prolific. But it's one of those things that depends entirely on your personal semver philosophy, like if you just live with whatever's released even if it was a mistake (Orthodox Semver? that's probably too religious), then it should be major.

lib/linkify.js
revin: happy to rename it if you like, I just mirrored the actual property name. `linkify.re` is actually an object full of named regular expressions. `regexes`?

lib/linkify.js
```diff
@@ -0,0 +1,55 @@
...
+    // linkify-it's default behavior is to create matches against the
+    // registered schema prefixes when the character immediately preceding a
+    // potential match is non-alphanumeric; however, we also need to skip
+    // linkification when said character is punctuation. So here we rebuild the
+    // internal regular expressions the LinkifyIt instance will use.
+    var regexes = linkify.re
+    regexes.schema_test = RegExp(regexes.schema_test.source.replace(regexes.src_ZPCc, regexes.src_ZCc), 'i')
+    regexes.schema_search = RegExp(regexes.schema_search.source.replace(regexes.src_ZPCc, regexes.src_ZCc), 'ig')
+    regexes.pretest = RegExp('(' + regexes.schema_test.source + ')|(' + regexes.host_fuzzy_test.source + ')|@', 'i')
+  }
```
revin: `regexes` it is

ashleygwilliams:
@revin, i think this needs a rebase

revin:
@ashleygwilliams yep, rebased!

lib/linkify.js
```diff
@@ -0,0 +1,55 @@
...
+//
+module.exports = function (parser) {
+  if (parser) {
+    var linkify = parser.linkify
+
+    linkify.set({fuzzyLink: false}) // turn off auto-linking normal hostnames
+    linkify.add('//', null) // turn off protocol-relative links
+
+    // linkify everything hostnamey starting with 'www.'
+    parser.linkify.add('www.', {
```
puzrin: Interesting solution with `www.` as schema. Did not expected such use, but that should work.

revin: @puzrin surprised me too (both that it worked, and that github's linkification is so simple/incomplete)

puzrin: Yeah, it's really horrible.

I could suggest to remove `.md` from TLD + new version has fixes for `//test` etc. That should resolve the most of real problems. But i understand your needs, and probably restricting fuzzy links to `www.[....]` will be more convenient for you.

lib/linkify.js
```diff
@@ -0,0 +1,55 @@
...
+      validate: function (text, pos, self) {
+        var match = text.slice(pos).match(/^\S*/)[0]
```
puzrin: (www.google.com) - that will fail with your regex. It seems better to reassemble honest regex from all that mess in `lib/re.js`

revin: Good point, I have to admit I wasn't aware of linkification of `<these> [types] (of links)` until today when you pointed to the linkify-it issue. I'll have to modify this.

puzrin: https://github.com/markdown-it/linkify-it/blob/1.2.3/index.js#L50-L65 - take this one as reference and try something like
```js
self.re.my_fuzzy /* use unique name */ =  new RegExp(
          '^' + self.re.src_host_port_strict + self.re.src_path, 'i'
        );
```

I'm not 100%, but seems that should help.

lib/linkify.js
puzrin: https://github.com/markdown-it/linkify-it/commit/9ffac07bed5df4dacf85e14cd2e70786fe63aac8 i did not announced this as feature, but in linkify-it 1.2.4 `www.` was added to "pretest" regex. Now you can replace fuzzy rule directly, without schema hacks.

lib/linkify.js
```diff
@@ -0,0 +1,84 @@
...
+          match.url = 'http://' + match.url.slice(1)
+        }
+        match.schema = ''
+      }
+    }
+
+    parser.linkify.add('www.', scheme)
+    parser.linkify.add('[www.', scheme)
+    parser.linkify.add('(www.', scheme)
+
```
revin: I sort of feel like specifically adding "schemes" for the bracket- and parenthesis-enclosed versions of these is cheating, but it's the most convenient way of allowing just those characters through while also blocking other forms of punctuation.

PR #152, opened 2016-02-10
Disable linkification of raw .md/.name domains/hostnames

revin:
This change disables `.md` and `.name` domains/hostnames from being linkified by markdown-it's linkification code. Other than those two, it creates links from all other hosts under valid TLDs, same as before.

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -54,6 +54,10 @@ module.exports = function (html, options) {
     .use(codeWrap)
     .use(expandTabs, {tabWidth: 4})
     .use(githubHeadings, options)
+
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
am hopping into a meeting but will take a look right after

ashleygwilliams:
:space_invader: :dolphin: :sparkling_heart: thank you thank you thank you

revin:
If I may, I'd suggest holding off on this in favor of #151.

ashleygwilliams:
closing in favor of #151 :+1:

PR #154, opened 2016-02-13
Remove extraneous code block wrapper customization code

revin:
The highlighted code block wrapper plugin we wrote for markdown-it had the beginnings of support for customizing the wrapper tag, but there was no way to change it from consuming code, and no tests that covered using a custom tag. Since doing it would force users to skip the HTML sanitizing step (because anything other than a `div` will get filtered out), it seems just as well to remove that stuff for now.


ashleygwilliams:
LGTM! :+1:

PR #158, opened 2016-03-01
hold github-url-to-object at 2.1.0

ashleygwilliams:
til we get #157 fixed. r? @revin


revin:
good thinking

ashleygwilliams:
@revin do you want to merge once travis is done? (i assume it should be all green :green_apple: )

zeke:
Sorry about that. Working with collaborators to get that regressions fixed. I will report back here when it's ready.

Issue #160, opened 2016-03-01
add test to show Mustache-style tags within README code blocks work


ashleygwilliams:
based on a previous issue, let's write a test to show that we *are* doing the right thing
------
https://github.com/npm/newww/issues/1750 from @passcod 

It appears that the Markdown renderer incorrectly strips out Mustache-style double-curlies in code blocks. Compare:

### Markdown source
https://raw.githubusercontent.com/Marak/faker.js/master/Readme.md

```
``` js
console.log(faker.fake('{{name.lastName}}, {{name.firstName}} {{name.suffix}}'));
// outputs: "Marks, Dean Sr."
```
```

(Using [Greek varias](https://en.wikipedia.org/wiki/Ancient_Greek_accent) to replace backticks there.)

### GIthub readme
https://github.com/Marak/Faker.js

![screenshot from 2016-01-15 10-38-53](https://cloud.githubusercontent.com/assets/155787/12338553/0a1dc046-bb75-11e5-8f54-2fe4428f7857.png)

### NPM readme
https://www.npmjs.com/package/faker

![screenshot from 2016-01-15 10-39-23](https://cloud.githubusercontent.com/assets/155787/12338571/17405324-bb75-11e5-8d17-87db4ccf7b4c.png)

revin:
I've looked at this one before; seems like a newww problem with the template rendering...?

Running the markdown above through marky via CLI renders it as:

```html
<h1 id="user-content-test" class="deep-link"><a href="#test">test</a></h1>
[...]

PR #161, opened 2016-03-01
Add tests to prove we aren't doing mustache template variable replacement

revin:
This should fulfill the request from #160


ashleygwilliams:
looks awesome.

Issue #162, opened 2016-03-01
Remove div.highlight wrapper if `highlightSyntax: false`


revin:
While working on #161, I discovered that when we call marky with `{highlightSyntax: false}` and it renders a fenced code block, we still have our `<pre>` element wrapped in `<div class="highlight [lang]">`. It seems like the correct behavior in this case to skip the wrapping altogether and simply output a `<pre>`. What do you think, @ashleygwilliams?

revin:
done in #163.

PR #163, opened 2016-03-01
Skip wrapping fenced code blocks when highlightSyntax === false

revin:
Hey @ashleygwilliams if you agree with the premise from #162, then here's code that omits the fenced code block wrapper when syntax highlighting is off.

(I assume this would go in at the next major version bump)


ashleygwilliams:
wonder if this will cause styling issues for the npm site. thoughts @aredridel or @jefflembeck ?

revin:
Yeah, it might. Does the npm site ever use marky with syntax highlighting off?

Issue #166, opened 2016-03-13
Feature request: add support for task lists


amercier:
Hi, I remarked [GitHub's Markdown task lists](https://github.com/blog/1825-task-lists-in-all-markdown-documents) are not properly rendering on https://www.npmjs.com/:

#### Expected

[![screen shot 2016-03-12 at 8 42 45 pm](https://cloud.githubusercontent.com/assets/1246795/13726958/188611ba-e893-11e5-9851-947e8406d566.png)](https://github.com/amercier/npm-package-skeleton)

Ex: https://github.com/amercier/npm-package-skeleton

#### Actual

[![screen shot 2016-03-12 at 8 45 40 pm](https://cloud.githubusercontent.com/assets/1246795/13726967/65e249d8-e893-11e5-9448-0efbc043a11f.png)
)](https://www.npmjs.com/package/package-skeleton)

Ex: https://www.npmjs.com/package/package-skeleton

It would be amazing if this feature was supported on npmjs.com! What do you guys think?

revin:
Hi @amercier, thanks for the request! I didn't want to let this sit here without a response, but I'm not the one to make a final decision on it; @ashleygwilliams is my boss on this repo :smile:, so it's up to her in the end.

That said, I personally don't see there being a major problem with supporting this, but I am in the middle of overhauling the way we parse markdown lists in general, so I probably can't do it until after that's finished.

For my own future reference, the task is to perform the following rendering on list items:

| leading | renders |
|---|---|
|`[ ]` | `<input class="task-list-item-checkbox" disabled="" type="checkbox">` |
|`[x]` | `<input class="task-list-item-checkbox" checked="" disabled="" type="checkbox">` |

revin:
fixed in #168.

Issue #149, opened 2016-02-09
Investigate switching from cheerio to markdown-it plugins for processing


revin:
A couple of our recent changes (#127, #133) have been solved by creating plugins for markdown-it to accomplish their processing steps. The idea popped into my head that it might be worth looking at moving the rest of our README processing into markdown-it the same way.

### Reasons to do this

Intuitively, it seems like this would result in greater speed and lower resource usage. **This is purely a hunch with no data to back it up.** The current process is roughly:

1. Render markdown to an HTML string using markdown-it
1. (optionally) Sanitize the HTML string
1. Parse the HTML string into a cheerio DOM object
1. Apply a series of transformations to the DOM object's contents
1. Return the DOM object to our calling code

Most of the time, I'm guessing marky's calling code simply calls `.html()` on the returned DOM object to get a string representation of the rendered page. For example, [npm/newww always does this](https://github.com/npm/newww/search?utf8=%E2%9C%93&q=marky&type=Code). If that's the case, maybe some future version of marky could process everything as part of the markdown-to-HTML rendering process and just return the string directly, skipping the cheerio-driven part altogether.

### Reasons *not* to do this

The main reason to avoid this is that we technically lose functionality if we do the straightforward version of this idea. The markdown-it processing code doesn't do anything with inline HTML; it just passes the markup through to the final rendered string. If we apply our transformations to the markdown text as it's being tokenized or rendered by markdown-it, we'll skip processing any inline HTML that the cheerio version of the process currently handles. That might be too risky to allow.

### Both?

It *might* be possible to create a markdown-it plugin that also parses inline HTML enough to create an object structure internally that our other plugins will still operate on. Then we'd get some of the benefits from each: efficiency due to leaner processing, but still the same rendered results.

All of this assumes that the cheerio DOM stuff is some kind of bottleneck, but again I have no idea if that's the case. This is purely speculative, but since @ashleygwilliams [recommended](https://github.com/npm/marky-markdown/pull/133#issuecomment-181534996) we have a separate issue here for the discussion, ... you're looking at it.

PR #168, opened 2016-03-24
GitHub-style task lists

revin:
I'd intended to add this as part of the work I'm doing on the overall list parsing, but it's enough of a standalone thing that I decided to make it a separate PR since I'm overhauling that other work (again). In case you decide the request from #166 is worth implementing, here you go: now we take leading `[ ]` and `[x]` in list items and convert them into disabled checkboxes, same as GitHub does, as described at:

* [Task Lists in GFM: Issues/Pulls, Comments](https://github.com/blog/1375-task-lists-in-gfm-issues-pulls-comments)
* [Task lists in all markdown documents](https://github.com/blog/1825-task-lists-in-all-markdown-documents)

_Note: this is rebased on top of #164 for ease of merging later_


zeke:
Love the idea. Would it make more sense as a `markdown-it` plugin, responsibility-wise?

revin:
@zeke it is! It's just not in its own repo. Or is that what you meant, that it would be better if we maintained this one separately?

zeke:
Yeah I meant a separate repo/package. Could be useful outside of marky-markdown town!

revin:
Yeah that's a possibility. If we end up going all the way with issue #149 and turning _everything_ into markdown-it plugins, maybe it all should be split out as it happens? PR #164 is sort of in the same place. Marky would grow super tiny!

revin:
ok @zeke I went your way and published the new stuff to npm as [markdown-it-task-lists](https://www.npmjs.com/package/markdown-it-task-lists), and updated this PR to pull it in rather than implement everything directly. I was lazy with the tests, however; they're substantially similar (read: basically the same) between repos :stuck_out_tongue:

PR #181, opened 2016-05-20
Test for leaked variables

aredridel:


test/marky.js
```diff
@@ -79,3 +80,9 @@ describe('debug', function () {
     unhookIntercept()
   })
 })
+
+describe('after', function () {
+  it('does not leak', function() {
+    assert.equal(Object.keys(global), oldkeys);
```
revin: As of the [fix in `markdown-it-expand-tabs`](https://github.com/revin/markdown-it-expand-tabs/commit/5f1f1ad5de2fa40536a14f9f79e8767ccc69ba1a), this test passes if you change from `.equal()` to `.deepStrictEqual()` (since we're comparing the members of different Array objects here, `.equal()` always fails no matter what)

revin:
Might be general github n00bness, but I didn't know how to make edits to this branch since it was in @aredridel's fork, so I copied/edited in a new branch which absorbs this; see #186. Thanks for the test case!

PR #185, opened 2016-05-24
Skip blank dependency readmes

revin:
This should fix #184, which was failing because the `mocha` devDependency didn't have a `README.*` and it was breaking the parsing in our tests.


revin:
eh you know what we don't need a separate PR here; I'll just leave the commit in the greenkeeper branch

PR #186, opened 2016-05-24
Test leak variable

revin:
This takes #181 (thanks, @aredridel!) and rebases it onto the changes from #184 because the tests don't work at all until that's merged. It also makes the [`.equal()`/`.deepStrictEqual()` change](https://github.com/npm/marky-markdown/pull/181#discussion_r64114887) I talked about.


revin:
Darn, failing on Travis in Node 0.1 and 0.12 due to lack of `.deepStrictEqual()`.

PR #191, opened 2016-05-28
8.0.0

revin:
This is to track the changes we make for the 8.0 release; we'll also need a release tag, no?

As of this writing, the merge order should be:

1. #151
2. #153
3. #163
4. #164
5. #168
6. #170
7. #203
8. This one.

_**Note:** all I'm doing in this branch is updating the changelog and the package version in `package.json`, but since there's a dependency ordering requirement 164 :arrow_right: 168 :arrow_right: 170, this PR is rebased onto 170, so all of their commits currently show up here._

_**Second Note:** this PR is just for tracking the changes; I'm not necessarily suggesting that these are the **only** things that should go into 8.0._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,100 @@
+# 8.0.0 (2016-07-08)
+
+Progress toward feature parity with GitHub's markdown rendering continues...
+
+### New Features / Breaking Changes
+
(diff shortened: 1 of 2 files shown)


zeke:
Nice work, @revin

PR #201, opened 2016-07-06
Fix heading link generation when inline HTML links exist

### LINKED ISSUE #200: Headers wrapped in links even if it already contains them.
The change at https://github.com/npm/marky-markdown/commit/569d0256b517f2135c03a3d1a7abe939b1ceef24 introduced an issue where headers with links may still be wrapped if they were html links. See https://github.com/npm/marky-markdown/commit/569d0256b517f2135c03a3d1a7abe939b1ceef24#commitcomment-18103230.

revin:
Should fix #200. We don't touch the inline HTML link, so there won't be any generated link that points to the slug generated for the heading's `id` attribute, but we won't double-wrap anymore, so that seems like a good tradeoff.

### DIFF
--- a/lib/headings.js
+++ b/lib/headings.js
@@ -2,6 +2,14 @@ var cheerio = require('cheerio')
 var GithubSlugger = require('github-slugger')
 var tokenUtil = require('./token-util')
 
+function isMarkdownLink (token) {
+  return token.type === 'link_open'
+}
+
+function isHtmlLink (token) {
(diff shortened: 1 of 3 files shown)


revin:
@ashleygwilliams think this can go in a patch release? The argument against doing so would be that someone somewhere is relying on the existing (fairly new) double-wrapping behavior @jdalton describes in #200.

I'm inclined to put it in 7.0.2, but that's a very weakly held opinion

ashleygwilliams:
imma say patch is good. this is a bug fix ;)

PR #203, opened 2016-07-07
Dont require 'highlights' in browserify environment.

### LINKED ISSUE #46: Browserify SyntaxError: Unexpected character '�' (2:2)
When I try and use Marky-Markdown with Browserify I get the following error

```
>> SyntaxError: Unexpected character '�' (2:2) while parsing C:\Users\noahh_000
Noah\Programing\Polymer\Polymer-Markdown-Tag\node_modules\marky-markdown\node_m
dules\highlights\node_modules\first-mate\node_modules\oniguruma\build\Release\o
ig_scanner.node while parsing file: C:\Users\noahh_000\Noah\Programing\Polymer\
olymer-Markdown-Tag\node_modules\marky-markdown\node_modules\highlights\node_mo
ules\first-mate\node_modules\oniguruma\build\Release\onig_scanner.node
```

This appears to be an issue with a dependency of Marky-Markdown.  

**OS**: *Windows 8.1*

wmhilton:
This change should make it possible to use marky-markdown with
browserify as long as the {highlightSyntax: false} option is used.
Using with {highlightSyntax: true} would generate a runtime error
in the browser.

Example command:
  browserify main.js -i highlights > bundle.js

Also note that by not including the highlights library, the file
size of the resulting bundle is reduced by half.

Would close #46

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,30 @@ npm i -g marky-markdown
 marky-markdown some.md > some.html
 ```
 
+## In the Browser
+
+This module mostly works in the browser, with the exception of the `highlights` module.
+You can compile it to a standalone file using `browserify`.
+
+```shell
+git clone https://github.com/npm/marky-markdown
+browserify marky-markdown/index.js -i highlights -s markyMarkdown > marky-markdown.js
+```
+
+Here is an example using HTML5 to render text inside <marky-markdown> tags.
+
+```html
+<script src="marky-markdown.js"></script>
+
+<marky-markdown>**Here** _is_ some [Markdown](https://github.com/)</marky-markdown>
+
+<script>
+  for (el of document.getElementsByTagName('marky-markdown')) {
+    el.innerHTML = markyMarkdown(el.innerText, {highlightSyntax: false}).html()
+  }
+</script>
+```
+
 ## Tests
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ git clone https://github.com/npm/marky-markdown
 browserify marky-markdown/index.js -i highlights -s markyMarkdown > marky-markdown.js
 ```
 
-Here is an example using HTML5 to render text inside <marky-markdown> tags.
+Here is an example using HTML5 to render text inside `<marky-markdown>` tags.
 
 ```html
 <script src="marky-markdown.js"></script>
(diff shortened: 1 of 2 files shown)


revin:
@wmhilton Interesting, thanks so much for looking into this! I like where this is going; is there a decent way to provide unit tests for the `process.browser != 'undefined'` case (i.e., in browserify)?

wmhilton:
That's a very good question! Would the goal be to test that simply running "browserify" on it doesn't error, or to actually to run the other tests on the "browserified" version? I wasn't imagining that browserify support would be officially supported... I just felt like this was a simple enough change to push upstream so I wouldn't have to manually patch marky-markdown.

wmhilton:
Holy bologna batman! It's already merged into master!? :D Thanks guys, that was quick. Glad I fixed that memory leak before I made the pull request... That's an awesome feature for a test suite btw; I gotta learn how that works.

revin:
@wmhilton yep, you caught us at a good time, since we're rolling out 8.0 shortly (gonna try and squeeze a few things in before it's published for real). 👍 

@ashleygwilliams do you have a preference on the testing question? The normal test suite just runs all the code in the newly introduced `if()` condition, so it doesn't even affect our code coverage reports at all. I'm inclined to agree with the comment above and leave it unsupported, but then the changelog does specifically call out the new feature, which makes it sound more official than anything. Should we just add a note clarifying this, perhaps?

revin:
@wmhilton btw if you're talking about the global variable leak test we have, that was all @aredridel, and yes, super handy—it led to a bug fix in one of our dependencies earlier!

(oh, also, I know you totally meant no offense, but we're not all "guys" here; "folks" maybe? 😄 )

wmhilton:
@revin Duly noted. I meant guys in the most inclusive sense of the word: men, women, people in between, robots, plants, animals, minerals. "Guys" is actually gender neutral in dictionary definition 2 (http://www.dictionary.com/browse/guy). However, for historical reasons, and since the existence of the word "gals" seems to diminish how gender-neutral the word "guys" is (and "guys & gals" reinforces the binary gender notion, making matters even worse) I'll try to adapt my lexicon, folks. :+1:

zeke:
This makes me so happy. 😂

PR #206, opened 2016-07-08
Bump highlights to 1.4.0

revin:
This ends up changing how syntax highlighted JS code renders, so it requires a test update.

**Note** that as of this writing, `highlights@1.4.0` isn't actually out yet, just `1.4.0-candidate`, for testing purposes. Once the final one is published, we can merge this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ Progress toward feature parity with GitHub's markdown rendering continues...
 
 ### New Features / Breaking Changes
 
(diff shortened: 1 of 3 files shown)


ashleygwilliams:
do we want to wait on this for 8.0.0?

revin:
IMHO, I'd like 8.0 to wait for this, just so we don't have to ship a 9.0 that's basically _only_ ES2015 highlighting. But I don't really have the same business pressures as everyone else involved, so I'll go with whatever you decide @ashleygwilliams.

PR #211, opened 2016-07-14
Browserify

zeke:
Following up on @wmhilton's work in https://github.com/npm/marky-markdown/pull/203, this PR: 

- Adds a browserify build script
- Includes a browserified file for distribution
- Uses the [browser field](https://github.com/defunctzombie/package-browser-field-spec) in package.json for easy `require`-ing
- Ensures the build is run before publishing
- Updates the "In the Browser" section of the README


ashleygwilliams:
restarting travis because i can't imagine how any of this would be breaking.

also hey! thanks for this @zeke !! will take a look when i find a sec :)

ashleygwilliams:
would also love to hear from @wmhilton !

package.json
```diff
@@ -3,8 +3,11 @@
   "version": "8.0.0",
   "description": "npm's markdown parser",
   "main": "index.js",
+  "browser": "dist/marky-markdown.js",
   "scripts": {
-    "test": "standard-format -w && standard && mocha --timeout 5000"
+    "build": "browserify index.js -i highlights -s markyMarkdown > dist/marky-markdown.js",
+    "test": "standard-format -w && standard && mocha --timeout 5000",
+    "prepublish": "build"
```
revin: Travis is complaining that it can't find the `build` command. `npm run build` here?

zeke: fixed!

revin:
oops now the `dist` file isn't passing the `standard` checker _[giant overdramatic sigh]_

package.json
```diff
@@ -3,8 +3,11 @@
   "version": "8.0.0",
   "description": "npm's markdown parser",
   "main": "index.js",
+  "browser": "dist/marky-markdown.js",
```
wmhilton: This is breaking things somehow. Investigating.

wmhilton:
Something is very wrong. This is what I get on master:

    $ browserify -i highlights -s markyMarkdown index.js | wc -l
    46991

This is what I'm seeing in this branch

    $ browserify -i highlights -s markyMarkdown index.js | wc -l
    8

zeke:
I see that too, @wmhilton. What the heck?

wmhilton:
I've figured out that removing the "browser" field in the package.json magically fixes it, but I don't understand why or how to do it properly yet. Too bad I'm not a browserify expert!

zeke:
Bizarre. I guess browserify itself is looking at the field? Chicken/egg problem?

Anyway, I've removed the `browser` key from package.json and commited a new build.

zeke:
I forgot that for some historical reasons, the `prepublish` script is run after `npm install`, so I moved the build to `pretest` instead.

wmhilton:
OK I think I get it now. Browserify (and now Webpack *cough* yes I might have that working too now, no promises!) prefer the raw source form rather than a precompiled form, so they can do their own minifying / deduplicating / shimming / whatever magic they do. So the "browser" field is not intended for a pre-built standalone, but as an alternative entry point. Instead of instead of having code that checks "process.browser" at runtime, you can have an alternate implementation. In fact, the browser field can be an object mapping files or dependencies to alternates or even telling browserify to ignore them.

What I've got, which seems to work, is:
```js
"browser": {
  "highlights": false
},
```

With that setting in the package.json, both browserify and webpack can "require" the module un-altered in code that is already using browserify:
```js
var mmmmmarkdown = require('marky-markdown')
```

So now the question is, do y'all still want to include a standalone compiled version in "dist"? Or is it enough that "npm install marky-markdown" work with webpack / browserify and folk can make their own standalone version if desired?

zeke:
> do y'all still want to include a standalone compiled version

I personally wouldn't use the built version, but it's a nice thing to offer for people who are not yet drinking the browserify Kool Aid™

wmhilton:
Alright! I give this one my seal of approval. It works with both browserify and webpack. With webpack it requires the json-loader module. <strike>I don't know how webpack handles such dependencies</strike> I looked into how and why webpack does NOT handle such dependencies and [it is a mess](https://github.com/webpack/webpack/issues/378). Short answer is webpack will inform the user that it needs a "json" loader and it is in their hands to choose an appropriate "json" loader, not the module author's. So... require('marky-markdown') with webpack will not work out-of-the-box, and that is working as intended. :confused: 

Here's my webpack command:
```shell
webpack --module-bind 'json=json' index.js bundle.js
```

I think... either just assume webpack users will figure this out, or put a note in the README: 

    Usage with webpack requires that your webpack.config.js configure a loader (such as json-loader) for .json files.

ashleygwilliams:
i am just watching and this all just makes me very happy 💖 👻 👾 🐠 💖

wmhilton:
Webpack apparently respects the ` "browser": {"highlights": false},` magic. I meant to say that earlier. :)

ashleygwilliams:
hey @zeke thanks so much for this PR. is it Ready-To-Go or you still working? if you'd like a review- could you squash and update the desc to reflect that convos that happened here? thanks so much!

zeke:
squashed and ready!

README.md
```diff
@@ -111,12 +111,9 @@ marky-markdown some.md > some.html
 ## In the Browser
 
 This module mostly works in the browser, with the exception of the `highlights` module.
-You can compile it to a standalone file using `browserify`.
 
-```shell
-git clone https://github.com/npm/marky-markdown
-browserify marky-markdown/index.js -i highlights -s markyMarkdown > marky-markdown.js
-```
+You can `require('marky-markdown')` in scripts you browserify yourself,
```
ashleygwilliams: let's link to browserify here :)

README.md
```diff
@@ -111,12 +111,9 @@ marky-markdown some.md > some.html
...
+or just use the standalone file in [dist/marky-markdown.js].
```
ashleygwilliams: code block please

README.md
```diff
@@ -148,7 +149,7 @@ npm test
 - [highlights-tokens](https://github.com/zeke/highlights-tokens): A list of the language tokens used by the Atom.app [highlights](https://www.npmjs.com/package/highlights) syntax highlighter
 - [lodash](https://github.com/lodash/lodash): A utility library delivering consistency, customization, performance, &amp; extras.
 - [markdown-it](https://github.com/markdown-it/markdown-it): Markdown-it - modern pluggable markdown parser.
-- [markdown-it-emoji](https://github.com/markdown-it/markdown-it-emoji): Markdown-it-emoji extension for Markdown-it that parses markdown emoji syntax to unicode. 
+- [markdown-it-emoji](https://github.com/markdown-it/markdown-it-emoji): Markdown-it-emoji extension for Markdown-it that parses markdown emoji syntax to unicode.
```
ashleygwilliams: can we put this white space change somewhere else (new commit/PR)? thanks!

ashleygwilliams:
ok! some very small nits- otherwise i think we are good to merge :)

ashleygwilliams:
actually, the nits are kinda stupid. i'm just gonna merge :)

PR #216, opened 2016-08-01
Allow text-align in style attributes on td/th elements, fixes #212

### LINKED ISSUE #212: Cell alignment is being stripped out from markdown tables
From [npm/newww#1933](https://github.com/npm/newww/issues/1933).

Tables' alignment specifiers (`:---`, `---:`, etc...) are being ignored. What's happening is, our sanitizer module is removing the `style` attributes that markdown-it is adding to the rendered `<th>` and `<td>` elements.

Simply adding `style` to the attribute whitelist would probably work, but that would allow everything. A more robust solution is to probably add logic to the [`exclusiveFilter()`](https://github.com/npm/marky-markdown/blob/v8.0.0/lib/sanitize.js#L64-L77) function that does more laser specific filtering on `th`/`td` `style` attribute contents.

revin:
Contrary to my guess in #212, the solution to allowing markdown-it's `text-align` table cell styles to make it through the sanitizer was not to update `exclusiveFilter()`, because that only allows a pass/fail style of operation. We also needed to potentially modify the `style` attributes as we process them, or else any inline HTML tables that have `text-align` in their `style` would always have a free pass to embed whatever styles they wanted. So instead we use sanitize-html's `transformTags` feature to do more precise filtering.


lib/sanitize.js
```diff
@@ -73,5 +73,41 @@ sanitizer.config = {
...
+  // generates `text-align` styles without whitespace
+  var style = attribs.style || ''
+  if (style.indexOf('text-align:left') >= 0) {
+    return cell('left')
+  } else if (style.indexOf('text-align:right') >= 0) {
+    return cell('right')
+  } else if (style.indexOf('text-align:center') >= 0) {
+    return cell('center')
+  } else {
+    return cell()
```
revin: @ashleygwilliams as you can see here and in the unit tests, markdown-it always generates the `text-align` stuff with no whitespace between the colon and the property value, so I'm just checking literally for that. A potential downside of doing it this way is that inline or hand-written HTML tables with cell alignments that look like, e.g., `text-align: center` (i.e, with a space in the middle) will get stripped out even though they're basically benign.

I have a unit test that specifically calls attention to this, to make it clear that the behavior is intentional. Should this maybe be a regular expression match with optional whitespace in the middle?

ashleygwilliams:
sorry i'm slow on this review. taking a look now!

ashleygwilliams:
do we think this is a bug fix or a feature? i'm leaning bug fix...

lib/sanitize.js
ashleygwilliams: i think we should do the regexp to allow the white space, in the end it will be more robust, and more closely matching the github behavior. lemme know if you want any help with that

test/fixtures/dirty.md
```diff
@@ -41,3 +41,27 @@ some<sup>superscript</sup>
...
 </dl>
+
+
+| First | Second | Third |
+|---:|:--:|:---|
+| Thing one | To be, or not to be, that is the question    | Hamlet     |
+| Thing two | Whether 'tis nobler in the mind to suffer    | Act 3      |
+| Row 3     | The slings and arrows of outrageous fortune, | scene 1    |
+| Row G     | Or to take arms against a sea of troubles    | 55–87      |
+| Row ZZZ   | And by opposing end them.                    | Chupacabra |
```
ashleygwilliams: lol love this example haha

ashleygwilliams:
yeah this is awesome, and ready to go once we add the allowing a space in the style attr-- unless you want to do that in another PR, which would also be fine. lemme know!

revin:
Updated w/the regular expression version. And yes @ashleygwilliams to answer your question above, I was thinking bug fix on this one; that makes the most sense to me.

test/sanitize.js
```diff
@@ -98,4 +98,33 @@ describe('sanitize', function () {
...
+  })
+
+  it('strips non-alignment table cell style', function () {
+    var html = marky(fixtures.dirty).html()
+    assert(!~html.indexOf('color:red;'))
+    assert(!~html.indexOf('width: 100%;'))
+
+    // alignment directives generated by something other than markdown-it's
+    // parser might contain internal whitespace; we normalize it such that our
+    // sanitized output contains none
```
ashleygwilliams: nice.

ashleygwilliams:
this looks awesome! merging.

PR #223, opened 2016-08-08
add an npm ignore

### LINKED ISSUE #222: ignore test files when publishing to npm?
this occurred to me since i'm going to build a little landing page for marky. shall we ignore the tests on publish? thoughts @revin ?

ashleygwilliams:
fixes #222 

still very much open to discussion, but while i had the thought, i figured i'd stand this up

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,3 @@
+test
+.travis.yml
+example.js


revin:
Good idea 👍

ashleygwilliams:
what semver should this be? i kinda feel like major but am not sure

revin:
oh, good question. I feel like it's one of those that's basically just bookkeeping/tidy work that doesn't feel "big" but still should be major because of what it's doing.

cmckni3:
I would consider this a patch.

Also, I think you should `npmignore` the test directory.

ashleygwilliams:
thanks for your feedback @cmckni3 ! we tend to be very conservative about semver, and since we have a major version release coming out, it'll probably just get lumped in with that!

cmckni3:
No problem. That's understandable.

Typically my development cycle is really fast, however, changes that aren't pressing like this sometimes get released with the next minor or major release.

ashleygwilliams:
ok i'm happy with this. merging.

PR #225, opened 2016-08-10
implement github style link headers

### LINKED ISSUE #224: Linking to headers that contain links
Is it possible, on npm, to link to a header that itself contains a link? For example,

```markdown
## Examples
```

lets you use `#examples` as part of a URL. On the other hand,

```markdown
## [Examples](https://github.com/nwhetsell/csound-api/tree/master/examples)
```

doesn’t let you use `#examples` as part of a URL on npm; https://www.npmjs.com/package/csound-api#examples doesn’t go to the Examples section of csound-api on npm, but https://github.com/nwhetsell/csound-api#examples does on GitHub.

Here’s a short script showing what marky-markdown produces in these cases:

```javascript
const marky = require('marky-markdown');

let HTML = marky('## Examples').html();
console.log(`==> ${HTML}`);
// ==> <h2 id="user-content-examples" class="deep-link"><a href="#examples">Examples</a></h2>

HTML = marky('## [Examples](https://github.com/nwhetsell/csound-api/tree/master/examples)').html();
console.log(`==> ${HTML}`);
// ==> <h2><a href="https://github.com/nwhetsell/csound-api/tree/master/examples">Examples</a></h2>
```

ashleygwilliams:
fixes #224 

to do:

- [x] update tests
- [x] clean up `util`

### DIFF
--- a/lib/headings.js
+++ b/lib/headings.js
@@ -2,6 +2,9 @@ var cheerio = require('cheerio')
 var GithubSlugger = require('github-slugger')
 var tokenUtil = require('./token-util')
 
+// shamelessy borrowed from GitHub, thanks y'all
+var svgLinkIconText = '<svg aria-hidden="true" class="deep-link-icon" height="16" version="1.1" viewBox="0 0 16 16" width="16"><path d="M4 9h1v1H4c-1.5 0-3-1.69-3-3.5S2.55 3 4 3h4c1.45 0 3 1.69 3 3.5 0 1.41-.91 2.72-2 3.25V8.59c.58-.45 1-1.27 1-2.09C10 5.22 8.98 4 8 4H4c-.98 0-2 1.22-2 2.5S3 9 4 9zm9-3h-1v1h1c1 0 2 1.22 2 2.5S13.98 12 13 12H9c-.98 0-2-1.22-2-2.5 0-.83.42-1.64 1-2.09V6.25c-1.09.53-2 1.84-2 3.25C6 11.31 7.55 13 9 13h4c1.45 0 3-1.69 3-3.5S14.5 6 13 6z"></path></svg>'
+
 var headings = module.exports = function (md, options) {
   if (options && !options.prefixHeadingIds) {
     headings.prefix = ''
@@ -37,8 +40,8 @@ var headings = module.exports = function (md, options) {
 
       // add 3 new token objects link_open, text, link_close
       var linkOpen = new Token('link_open', 'a', 1)
-      var text = new Token('text', '', 0)
-      text.content = 'link'
+      var text = new Token('html_inline', '', 0)
+      text.content = svgLinkIconText
       var linkClose = new Token('link_close', 'a', -1)
 
       // add some link attributes
--- a/lib/headings.js
+++ b/lib/headings.js
@@ -2,14 +2,6 @@ var cheerio = require('cheerio')
 var GithubSlugger = require('github-slugger')
 var tokenUtil = require('./token-util')
 
-function isMarkdownLink (token) {
-  return token.type === 'link_open'
-}
-
-function isHtmlLink (token) {
-  return token.type === 'html_inline' && token.content.substring(0, 3).toLowerCase() === '<a '
-}
-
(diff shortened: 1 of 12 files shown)


lib/headings.js
```diff
@@ -19,22 +19,27 @@ var headings = module.exports = function (md, options) {
...
   md.core.ruler.push('githubHeadings', function (state) {
     // save the Token constructor because we'll be building a few instances at render
     // time; that's sort of outside the intended markdown-it parsing sequence, but
     // since we have tight control over what we're creating (a link), we're safe
     if (!savedTokenConstructor) {
       tokenUtil.set(state.Token)
+      Token = state.Token
       savedTokenConstructor = true
     }
   })
```
revin: It occurred to me while doing something else that we can eliminate `savedTokenConstructor` from this file and just do
```js
if (!Token) {
  Token = state.Token
  tokenUtil.set(Token)
}
```

lib/headings.js
```diff
@@ -19,22 +19,27 @@ var headings = module.exports = function (md, options) {
...
   md.renderer.rules.heading_open = function (tokens, idx, opts, env, self) {
-    // Bail if heading already contains a hyperlink
+
+    
+
+
     var children = tokens[idx + 1].children
+    // make sure heading is not empty
     if (children && children.length) {
-      if (children.filter(isMarkdownLink).length === 0) {
```
revin: now we can remove `isMarkdownLink` from up above 🎉

lib/headings.js
```diff
@@ -44,15 +49,21 @@ var headings = module.exports = function (md, options) {
...
+        text.content = "link"
+        var link_close = new Token('link_close', 'a', -1)
+        
+        // add some link attributes
+        link_open.attrSet('id', headings.prefix + postfix)
+        link_open.attrSet('class', headings.className)
+        link_open.attrSet('href', "#" + postfix)
 
-        // if the heading doesn't already contain a link, wrap the contents in one
-        if (children.filter(isHtmlLink).length === 0) {
```
revin: now we can remove `isHtmlLink` from up above 🎉

lib/headings.js
```diff
@@ -44,15 +49,21 @@ var headings = module.exports = function (md, options) {
...
-          tokenUtil.wrap(children, postfix)
```
revin: also, so I don't forget, we'll be able to remove the `wrap` function from token-util.js now too.

delete all teh codez

ashleygwilliams:
I LOVE DELETING CODE

ashleygwilliams:
```
119 passing (4s)
  33 failing
```

revin:
33, wow. I guess that part of the rendering is well covered 😛

revin:
✨`152 passing` ✨

ashleygwilliams:
niiiiiiiice 👯

ashleygwilliams:
thought -> should the SVG icon be a flag? @revin

revin:
@ashleygwilliams probably, you're right. We have a slightly now misnamed option `prefixHeadingIds` that enables or disables the 'user-content-' string that gets prepended to the generated `<a>` ids, but nothing for skipping the heading processing completely.

Want an option that disables the link generation altogether? Or an option that enables it, default `true`, and you can set it to false manually if you want? Basically we would be allowing folks to completely skip our githubHeadings plugin completely.

Also since this is already semver-major, should we rename `prefixHeadingIds` to be more accurate? Something like ... `prefixHeadingLinkIds` ...? I don't know, seems long, and you could argue that conceptually the IDs _are_ for the headings, just not the `<h1>`, `<h2>`, etc. elements themselves. I think I just talked myself into leaving that alone.

revin:
Oh wait, did you mean disabling _just_ the icon itself, but still doing the rest of the processing (so we'd still generate the slug and link, but the link would be empty)?

ashleygwilliams:
hah yes, the second thing, just the SVG part. something like `--without-icons`

revin:
I added this as `options.enableHeadingLinkIcons` for programmatic usage; maybe we could add the CLI part in #132?

PR #235, opened 2016-08-16
Allow link/image titles

revin:
I learned that in link URLs and image paths, everything after a space is supposed to be interpreted as the `title` attribute, but our sanitizer was stripping that out.

GitHub supports it, as verified here by hovering your mouse over the rendered pieces

### Links

    [marky-markdown](https://www.npmjs.com/package/marky-markdown "an npm joint")

> [marky-markdown](https://www.npmjs.com/package/marky-markdown "an npm joint")

### Images

    Build Status

> Build Status

---
All we have to do is add them to the whitelists in the sanitizer module and add unit tests.


dist/marky-markdown.js
```diff
@@ -672,16 +672,16 @@ sanitizer.config = {
...
     meta: ['name', 'content'],
     iframe: ['src', 'frameborder', 'allowfullscreen'],
     input: ['checked', 'class', 'disabled', 'type'],
     div: ['id'],
     span: [],
     pre: [],
-    td: ['colspan', 'rowspan'],
-    th: ['colspan', 'rowspan'],
+    td: ['colspan', 'rowspan', 'style'],
+    th: ['colspan', 'rowspan', 'style'],
```
ashleygwilliams: why is this in here? wasn't this from https://github.com/npm/marky-markdown/pull/216/files? i'm confused because changes to `td` and `th` aren't in the desc for this. unless i'm missing something...

dist/marky-markdown.js
```diff
@@ -4773,17 +4793,16 @@ module.exports={
   "_location": "/cheerio",
   "_nodeVersion": "5.5.0",
   "_npmUser": {
-    "name": "feedic",
-    "email": "me@feedic.com"
+    "email": "me@feedic.com",
```
ashleygwilliams: any reason for this switch? (just trying to understand)

ashleygwilliams:
changes on lines 4773-4882 seem like they are being generated from something? i can't tell what's going on there-- do you @revin? 💁 ❓

revin:
Yeah I'm not sure about that, I haven't been paying much attention to the browserify bundle because it just gets built every time I run `npm test` and I've been assuming it'll get cleared up with a final test run after all the 9.0 PRs are merged.

Want me to remove it from the commits? The diffs do seem to be basically just clutter/misdirection.

ashleygwilliams:
it looks like a shrinkwrap file honestly... pinging @zeke -- any help on identifying this stuff?

ashleygwilliams:
i *think* we should remove it but  let's wait to hear what @zeke says

zeke:
Weird. I don't know what that's about.

Possible ways to work around the annoyance of a frequently-changing dist file:

1. Only run the build script before a release, instead of in a `pretest` script
1. gitignore the file, but don't npmignore it.

revin:
OK I split the browserify bundle out of this PR, so it's just the code that changed

ashleygwilliams:
🐬 👻 ✨

PR #234, opened 2016-08-15
GitHub-style (non-)parsing of indented headings

revin:
This stops marky from parsing ATX (i.e., single line `#`-style) markup as headings if there's leading whitespace before the `#` character. It's not CommonMark, but it's how GitHub does it.

Because of the similarity in internal module naming going on, I've rebased this on top of #225, so merge that first. This PR is really just one commit.

Fixes #233.

PR #237, opened 2016-08-18
Indented headings

### LINKED ISSUE #233: # with space preceding interpreted as heading
I was considering filing this with markdown-it but I haven't dove into how much customization marky does and since marky's goal is GitHub parity I am asking here first. 

In this [readme rendered by marky](http://render-readmes.preview.kitnic.it/boards/github.com/8BitMixtape/NextLevelEdition/#8bit-mixtape-nextlevel-shit) the linked title shouldn't be a title. [Here it is rendered on Github](https://github.com/8BitMixtape/NextLevelEdition#how-to-install). 

There is a space preceding the `#` so it seems GitHub doesn't consider it a title but marky does.

ashleygwilliams:
replacing #234 that i merged too early

from @revin: 
This stops marky from parsing ATX (i.e., single line #-style) markup as headings if there's leading whitespace before the # character. It's not CommonMark, but it's how GitHub does it.

Because of the similarity in internal module naming going on, I've rebased this on top of #225, so merge that first. This PR is really just one commit.

Fixes #233.


ashleygwilliams:
looks good to me! cc/@revin

PR #246, opened 2016-09-21
Replace cheerio-based heading ID processing with custom string processor

revin:
Hey @ashleygwilliams please look this over first, but I think this works for taking cheerio out of the process by which we generate heading link IDs.

I have a local cache of the entire npm package repository's README files that I pulled a few days ago using a modified readme-tester, and I ran marky against the first 100,000 files (which come in sorted alphabetically by package name). The difference before/after this change is small, but it looks like 0.5–1ms per readme on average, so it's something. It's not remotely enough for #149 to be done, but it's a step in the right direction. I think?

### DIFF
--- a/lib/heading-links.js
+++ b/lib/heading-links.js
@@ -1,5 +1,5 @@
-var cheerio = require('cheerio')
 var GithubSlugger = require('github-slugger')
+var innertext = require('innertext')
 var tokenUtil = require('./token-util')
 
 // shamelessly borrowed from GitHub, thanks y'all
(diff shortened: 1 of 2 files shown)


lib/innertext.js
```diff
@@ -0,0 +1,20 @@
...
+function innertext (html) {
+  // Drop everything inside <...> (i.e., tags/elements), and keep the text.
+  // Unlike browser innerText, this removes newlines, but since we're running
+  // this against heading text generated with markdown-it, we don't expect this
+  // difference to cause problems.
+  return entities.decode(html.split(/<[^>]+>/)
+    .map(function (chunk) { return chunk.trim() })
+    .filter(function (trimmedChunk) { return trimmedChunk.length > 0 })
+    .join(' ')
+  )
```
revin: The biggest question I have is whether this logic is sufficient. It passes our unit tests, but it feels pretty heavy-handed and lacking in…grace, I guess? Like there must be example inputs that render this invalid, but I couldn't think of them.

ashleygwilliams:
@revin does this have a breaking change in it? i.e. does this break any links (to the best of your knowledge? i know it's hard to account for all the stuff, just your best guess works)

ashleygwilliams:
ok after looking this over a bunch i think this looks correct. i'd like to merge it and if there's some weirdness we can send issues/PRs to `innertext` :)

PR #251, opened 2016-09-29
Remove cheerio dependency by converting to markdown-it plugins

revin:
Here's an attempt at addressing #149. It's based on the current master, so if we want to put this in 10.0, I'll rebase it when we release 9.0; or we can push it all into this release if you want.

The main breaking change here, of course, is that marky will now return a string of HTML rather than a cheerio DOM object. This is easy enough to deal with; you just do `cheerio.load(marky(...))` now. This is what the test suite does now in most places (so we didn't get rid of cheerio _all_ the way; it's a devDependency here)

Performance-wise, I ran the first 100,000 readmes in a local cache I pulled from the registry a couple of weeks ago using a modified `readme-tester`. I didn't do anything to monitor memory consumption, but the speed gains are encouraging:

### Before

This is using the current master branch. I ran this as I left for the evening one day, so it was the only thing really running on my laptop.
```
{
  num: 100000,
  sum: 2021762.7870619856,
  avg: 20.217627870619857,
  min: 1.063849,
  q1: 5.3645795,
  median: 9.499271,
  q3: 20.900199999999998,
  max: 977.573982
}
```

### After

This is using the PR branch. I ran this during the day while I was doing other work, jumping in and out of video conferences, etc...
```
{
  num: 100000,
  sum: 1689549.6293209926,
  avg: 16.895496293209927,
  min: 2.56501,
  q1: 4.106909,
  median: 7.672472,
  q3: 17.0588025,
  max: 997.958522
}
```

I suspect the reason the`min` and `max` numbers are higher in the new numbers is because, as I said, I was doing some other work while it ran in the background. This isn't scientific; just ballpark number generation. But still...

**Overall it looks like we're a good 15-18% faster with this change, which seems to confirm the hunch that caused me to create #149 in the first place.**


ashleygwilliams:
does this have any effect on the API marky exposes? also 😍

revin:
@ashleygwilliams so far, everything is the same except for the return value (HTML string vs cheerio DOM object). The parsing/rendering options all work the same as before.

I moved the `marky.parsePackageDescription()` definition out and factored it slightly differently than before, but the tests around that all still pass, so I'm pretty confident that part still works. I remember seeing npm web site source that used it a long time ago, and I think it was basically just `require('marky-markdown').parsePackageDescription(...)`, so I'd only expect issues if that usage changed.

zeke:
I'm kinda sad to see cheerio factored out, but I can't totally justify that reaction. I found the cheerio thing most useful for writing the test suite, but in practice most people probably just end up calling `html()` on the result. If users still want a DOM object, it would now just be a `cheerio.load` away.

One reason not to ship this would be that the code may now be a bit harder to understand. This, for example, is a little dautning:

```js
return originalRule.call(this, tokens, idx, options, env, self)
```

Performance is important, but if I remember correctly the readmes are being cached in redis anyway, so they're probably not actually generated that often.

That said, I don't feel very strongly about this. Just wanted to leave my two cents on the matter.

revin:
Thanks @zeke, totally fair. It's absolutely true that code for cheerio-based operations is a lot clearer than code that manipulates the markdown-it token array. If everyone looks at this and says welp, that was a nice experiment but we think the clarity is worth the performance hit so let's skip this PR, I'm perfectly happy with that result.

ashleygwilliams:
i personally think the performance portion is huge. i'd like this to be the fast nodejs github-flavored markdown parser, not npm specific (note #257). i also think that reducing our dependencies and layers of manipulation is going to eventually make the code easier to reason about and extend.

i am a HUGE fan of marky being a custom markdown-it instance for which we just plugin what is necessary to achieve parity with github flavored markdown. it's a simpler, reasonable architecture which significantly reduces the need to think through any "marky-specific" logic.

i've been talking about this internally with the npm team and people are really excited. i'd be interested in what others think, but i also think that given our product goals this is right on point.

ashleygwilliams:
since this is a big change @wmhilton @woorm @aredridel @ceejbot @jefflembeck any thoughts?

wooorm:
I’m 👍 on dropping Cheerio, less big moving parts is more better, and I agree with the points @ashleygwilliams pointed out in https://github.com/npm/marky-markdown/pull/251#issuecomment-251735355!

revin:
@wooorm was looking at `is-badge` for this, and the marky readme has a badge from codeclimate.com and a couple from issuestats.com, but `isBadge(…)` returns `false` for those; should I file issues for them? (It's literally the only markdown file I've tested so far, so I don't know what other common ones aren't supported yet)

revin:
Marking this as WIP for the moment while I look into `is-badge`

wooorm:
👍

I added you both as a collaborator, so feel free to add more (or I’ll try to add them tomorrow, heading to bed). I just added codecov, in https://github.com/wooorm/is-badge/commit/a83f62b38de44d3c427729a919556577d10d1a57, so that should give an idea on what’s required for a new badge to be added.

zeke:
Didn't know about `is-badge`. Small modules! 🍰 

Also, this is the first time I'm seeing a warning from dependency-ci:

screen shot 2016-10-10 at 2 47 39 pm

Strangely, I'm not seeing a deprecation message for that package:

```
❯ npm i lodash.assign@'^4.0.2'
- lodash.keys@4.2.0 node_modules/lodash.keys
- lodash.rest@4.0.5 node_modules/lodash.rest
Desktop@1.0.0 /Users/zeke/Desktop
└── lodash.assign@4.2.0
```

cc @andrew

ashleygwilliams:
@zeke can u move that discussion to #254

revin:
Update: finished migrating this PR to use `is-badge` (thx @wooorm), and fixed a logic bug in the markdown-it badge plugin.

I'm not quite sure why dependency-ci is still flagging lodash.assign as deprecated 😕

andrew:
@revin I've just updated the info in Libraries.io as it missed the undeprecated-ness, can you try rerunnign the dependency ci build now?

revin:
Oh, fantastic, thanks @andrew that seemed to fix it 👏👏👏👏👏

revin:
Updated this to include what's in #270, but should probably rebase this after that or something, so it's not duplicated in the end

ashleygwilliams:
whoops looks like it's not so happy after i merged #270. 

also, i've been holding off on merging this cuz i am not entirely sure it's good. i *think* it looks good. anything i should be looking at specifically @revin ? if not, then i think we are good to merge, i'm just nervous, heh 😅

revin:
(I just now rebased it against the new master, so once it travises, we should be good to go)

@ashleygwilliams yeah I gotcha, totally understood. I don't really have any questions about any of this myself, other than just, "do we want to do this for sure?" per the code/clarity/speed tradeoffs discussed above. It passes our test suite just fine, so I don't have any concerns on that front.

Building it out _did_ clue me into a few of the areas that had kind of fallen into disrepair (like the badges, etc...), but it's all stuff you ended up filing issues on for 10.0, so we can deal with it there.

zeke:
> once it travises, we should be good to go

travis as a verb :+1:

ashleygwilliams:
OK. i'm doing it. because all other PRs really need this to merged. farewell cheerio!

ashleygwilliams:
omg @wmhilton 😂

PR #270, opened 2016-10-19
Allow `align` attribute on `<img>` elements

### LINKED ISSUE #269: support align attributes on images in markdown readme files
from https://github.com/npm/www/issues/51 by @revelt 

<hr>
Hi all,
You probably noticed that GitHub let's you float images on the README markup documents using `align="left|right"`, but npmjs.com does not. This is counter-intuitive since GitHub doesn't have the sidebar on the right, but npmjs.com does — readme's are even more cramped on this website! Supporting the `align=` attribute on `` tags would relieve this problem.
cheers,
Roy

revin:
Fixes #269. I'll do this in #251 as well, in which case this one will become redundant (but it's here in case we don't merge it for 9.0 for any reason)

### DIFF
--- a/lib/sanitize.js
+++ b/lib/sanitize.js
@@ -51,7 +51,7 @@ sanitizer.config = {
     h5: ['id'],
     h6: ['id'],
     a: ['href', 'id', 'name', 'target', 'title'],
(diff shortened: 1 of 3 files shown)


ashleygwilliams:
this looks great! mergin' 🏇

PR #271, opened 2016-10-20
Remove dist

### LINKED ISSUE #262: gitignore `dist.js`, still publish to registry
merge conflicts in this built file are a huge PITA. we shoudn't be commiting it anyways. let's fix this for 9.0.

karanjthakkar:
Fixes #262


ashleygwilliams:
🎉 

you might want to use `npm pack` to eventually test that it is being included in the tarball on publish

karanjthakkar:
@ashleygwilliams I used `irish-pub` to check the files being included in the package. All seems to be working for me. What I've done is, instead of using the `files` key in `package.json`, I whitelist `dist` in `.npmignore`. The reason I'm doing that is because the files array requires explicit mention of all the folders/files to be included. I wasnt sure if that was something you wanted. So I went ahead with the second option. Let me know what you think.

ashleygwilliams:
this looks good to me! any thoughts @revin? thanks @karanjthakkar !

ashleygwilliams:
yay! thank you so much! 👻 🐕 🐬

Issue #256, opened 2016-10-05
document Node version support


ashleygwilliams:
related:
- https://github.com/npm/marky-markdown/issues/240
- #255

ashleygwilliams:
this was fixed by #273

PR #273, opened 2016-10-22
update and document Node version support

### LINKED ISSUE #255: formally end support for 0.10
0.10 is EOL on Oct 31st. we should remove it from travis.

related: https://github.com/npm/marky-markdown/issues/240

ashleygwilliams:
fixes #255 and #256


revin:
Excellent, thanks @ashleygwilliams 😀

Issue #274, opened 2016-10-25
review youtube embed handling


revin:
Similar to #258 in intent, we have a [markdown-it plugin](https://github.com/npm/marky-markdown/blob/master/lib/plugin/youtube.js) that wraps iframes pointing to youtube videos in `<div class="youtube-video">…</div>`. However the sanitization stuff #259 links doesn't have `iframe` in the whitelist, so it appears that GitHub strips them completely. I couldn't find a readme that included a youtube embed to verify this, and testing in gist comments just now didn't show me anything to indicate otherwise.

I think we can remove this altogether as it's both unused and unnecessary for GitHub compatibility.

revin:
For anyone looking to make your first PR 😍 the task here is to remove the [plugin file](https://github.com/npm/marky-markdown/blob/30865e4c37fae402d9e18b83a669bf8103259d27/lib/plugin/youtube.js) and remove the usage of it from [render.js](https://github.com/npm/marky-markdown/blob/30865e4c37fae402d9e18b83a669bf8103259d27/lib/render.js).

ashleygwilliams:
iiinteresting. maybe i'll write something to see if we have anyone using youtube videos!

ashleygwilliams:
ok! so i wrote a small follower here: https://github.com/ashleygwilliams/youtube-readme-follower

the results so far seem to be for packages that help embed youtube videos. or link to a youtube video and then also talk about some other type of embedding.

it occurs to me that there are a few places that npm uses marky that aren't strictly for markdown parsing of readmes, namely:

- tips (https://github.com/npm/npm-website/blob/169eee7f39011a86ae8f39c2995ac567962e32bd/handlers/login.js#L79)
- on the docs (https://raw.githubusercontent.com/npm/docs/master/content/getting-started/installing-npm-packages-globally.md)

i don't think that `tips` on the website ever contains content with an iframe in it. the docs site definitely does. i wonder how we can handle such a thing.

it is *definitely* true that github crops em out tho.

i'll comment with the final results of the follower code when it finishes.

revin:
oh nice, thanks for writing a proper follower; I just used my weeks-old readme cache I built for doing performance tests, and then `$ ag 'iframe src=.*youtube.com' readmes/` which isn't very robust at all. Picked up 10 packages that, as you said, help people embed videos (so all of the iframes were actually in code snippets, not in the real docs).

shibasisp:
I am a complete newbie and I would like to work upon it..

[...]

PR #275, opened 2016-10-26
solved #274 review youtube embed handling

shibasisp:
is this okay?


shibasisp:
can anyone please tell me what is the problem? Why the tests are failing?

revin:
Hey @shibasisp, that was quick! Fantastic!

Guess what? I totally forgot to list that we have unit tests around this as well, and for TravisCI to pass the PR, the youtube unit tests need to be removed as well. Oops. Those are in the `test` directory.

_However_, @ashleygwilliams brought up a good point in #274, that the ideal outcome here might be different than just removing the youtube stuff, because there are definitely users (the npm site, at least) that need it. So if you don't mind, I have to hang on to this until she's able to investigate a bit more and we can decide where the youtube processing stuff needs to live.

shibasisp:
Ok fine..

ashleygwilliams:
hi! so yes- until i can completely move the npm docs off of marky-markdown (it's in the plan, just not a priority at the moment) we'll need to keep this functionality. that being said, because it is not inline with the product direction of parity with github-flavored markdown, i would be happy to see it moved behind a flag as part of the deprecation path! cc/@revin

revin:
That sounds good. So @shibasisp does that sound interesting to you? Basically add an option in index.js, say, `allowDeprecatedYoutubeEmbeds: false` (unless you have any particular preference for the name, @ashleygwilliams), and only `.use(youtube)` in render.js if the option is `true`, similar to how the syntax highlighting and CDN things work.

Then instead of removing the tests, they'd need updated so the current tests pass if the option is set to `true`, and add some new tests to verify that the `<iframe>`s are stripped out when the option is set to `false` or omitted.

Is that enough information for you to know what changes to make? I'm totally happy to help if anything is confusing. 👍

ashleygwilliams:
no preference on name for the flag!

revin:
Hey again @shibasisp, do you have the time/availability to implement the version of this where it's deprecated and not removed entirely? If not, I can take over from here if you like. No pressure. Thanks again!

Issue #217, opened 2016-08-02
links not wrapped in <p> when h3 is html.


jdalton:
I noticed that

```
### foo
[a](foo) [b](bar)
```

generates html like:

```
<h3>foo</h3>
<p>
<a href="foo">a</a> <a href="bar">b</a>
</p>
```

but

```
<h3 id="baz">foo</h3>
[a](foo) [b](bar)
```

generates html like:

```
<h3 id="baz">foo</h3>
<a href="foo">a</a> <a href="bar">b</a>
<p></p>
```

revin:
Hey @jdalton thanks for writing this up! 👍 

CommonMark has some interesting handling rules for HTML. The section on [HTML blocks](http://spec.commonmark.org/0.25/#html-blocks) describes the conditions in which some markup is considered an "HTML Block", and even though `<h3>` is a block-level tag, it's not in the list of tag names that create HTML Blocks, so it gets parsed as [raw HTML](http://spec.commonmark.org/0.25/#raw-html) instead, which means

    <h3 id="baz">foo</h3>
    [a](foo) [b](bar)

_should_ render as

    <p><h3 id="baz">foo</h3>
    <a href="foo">a</a> <a href="bar">b</a></p>

i.e., the wrapping `<p>` should wrap the entire thing, and it does, in the [markdown-it demo](https://markdown-it.github.io/). For some reason we have the empty trailing `<p>` instead. So that's strange.

However, it might be irrelevant, because this is an area where GitHub renders the perhaps more intuitive way it looks like you're expecting, so if we're going to change anything, it should probably be in that (non-standard) direction, yeah?

revin:
An interesting related thing I discovered while looking at this is that `<h1>` _is_ in the list of tag names that create HTML Blocks, so if it was an `<h1>` here rather than `<h3>`, the markdown is considered to be part of the HTML Block, and [doesn't get translated to HTML at all](http://spec.commonmark.org/0.25/#example-126), unless you put a blank line in between (the blank line terminates the HTML Block).

jamestautges:
#278 should fix this. I had to write a plugin for the markdown-it parser to circumvent it treating the header as inline html.

revin:
#278 is merged; closing

PR #278, opened 2016-10-26
Parse raw html headings like regular headings

jamestautges:
This change addresses issue #217.

Instead of treating raw header tags as html blocks or whatever it was doing in that issue, the new parsing rule wraps up the block just like a regular heading block.


revin:
oh wow, this is so cool @jamestautges, thank you!

It looks like this was based on the original markdown-it source, yes? Our test suite will fail if any of the code doesn't fit our formatting rules as enforced by `standard`, so before TravisCI will pass it, it has to be reformatted accordingly 😕

This is also something that would probably benefit from being backed by some unit tests anyway because it's a deviation from markdown-it's normal CommonMark compliant algorithm, so I'll mark this as 'needs tests' for the time being.

Please let us know if you want guidance on any of that; very happy to help (and to have your help)!

jamestautges:
@revin How's that?

revin:
@jamestautges oh wow that looks much improved, nice 😄 

When you run `npm test` is it picking up the new tests? I would've expected it needed a line in test/index.js or something (unless I'm incorrectly remembering how it works)

jamestautges:
@revin Oh yeah you're right. I was just testing the file directly. I'll fix it.

revin:
@jamestautges How does this interact with the plugin we have in lib/plugin/heading-links.js? If we have:
```markdown
<h3>Some HTML Heading</h3>
[a](foo) [b](bar)
```
does the rendered output look like:
```html
<h3><a id="user-content-some-html-heading" class="deep-link" href="#foo"></a>Some HTML Heading</h3>
<p><a href="foo">a</a> <a href="bar">b</a></p>
```
(or just `id="some-html-heading"` if `{prefixHeadingIds: false}` is in the options)

Based on a test I just ran on GitHub, they do the heading links irrespective of the way the headings were created, so we'll need to do that too.

revin:
To be clear, it looks like it should work just fine since you're creating the proper token types; maybe all we need here is a couple of unit tests (with and without `prefixHeadingIds` set) to verify.

revin:
The other thing I was going to mention was the GitHub behavior we replicate in lib/gfm/indented-headings.js, where `# foo` style headings are parsed as paragraphs rather than headings if they have any leading whitespace. It doesn't look like they care about whitespace appearing before headings created by inline HTML, so I think there's nothing to change in the parsing rule here, but would you mind creating a couple of tests to cover, e.g.,
```markdown
 <h2>Something with one space indent stays a heading</h2>
 paragraph text
```
```markdown
  <h3>Something with two spaces indent stays a heading</h3>
  paragraph text
```
```markdown
   <h4>Something with three spaces indent stays a heading</h4>
   paragraph text
```
```markdown
    <h3>With four spaces this should become a code block</h3>
    paragraph text
````
and of course if the indented stuff appears in the document after a list, GitHub will typically append it to the list. Verifying/enforcing that is probably outside the scope of this PR though, yeah?

(As you can see, we tend to favor being fairly thorough with testing corner cases like this where GitHub's parsing deviates from CommonMark)

If for some reason that last example still becomes a header rather than a code block like it should, I bet it would just take something like `if (state.tShift[startLine] > 3) { return false }` near the top of your rule so it bailed out from building the heading tokens.

Thanks so much for this, I'm excited to get it in! 🎉🎉🎉

ashleygwilliams:
checking in on progress here @revin @jamestautges lemme know if this needs anything from me to move forward!

jamestautges:
How does that look? I copied over all the tests I though were applicable to the new case and modified them.

revin:
Fantastic @jamestautges thanks so much! Merged it for 10.0.

ashleygwilliams:
🎉 thank you so much!! be on the lookout for the 10.0 release, we'll make sure you get a shoutout!

PR #279, opened 2016-10-26
Support spaces in path names

sjking:


revin:
oh wow, I wasn't expecting this one, awesome @sjking 👍 

I'm going to have to read through this carefully later, but the first thing I noticed is that it looks like me merging your other PR with the Rebase And Merge button changed the history enough that it's showing up here as well (as the original commit); could you please rebase this onto the current master? Should get rid of that.

@ashleygwilliams what do you think about the overall design here? I had originally planned on pulling in only the file(s) from markdown-it that needed to actually change, and then possibly having those pull in individual modules from markdown-it where needed, like `require('markdown-it/lib/some/utility/module')`. Would you rather duplicate all of the utility/helper code (advantage: we stay more resilient against internal markdown-it changes; disadvantage: more code to maintain), or reach inside markdown-it for whatever we can (advantage: less code to maintain here; disadvantage: we couple ourselves to markdown-it's internal design more)?

sjking:
@revin Please see the other pull request, this one got messed up 🤕 
PR #280

revin:
ok, will do!

PR #280, opened 2016-10-27
Support spaces in path names

sjking:


revin:
Importing this question from the other PR:

@ashleygwilliams what do you think about the overall design here? I had originally planned on pulling in only the file(s) from markdown-it that needed to actually change, and then possibly having those pull in individual modules from markdown-it where needed, like `require('markdown-it/lib/some/utility/module')`. Would you rather duplicate all of the utility/helper code (advantage: we stay more resilient against internal markdown-it changes; disadvantage: more code to maintain), or reach inside markdown-it for whatever we can (advantage: less code to maintain here; disadvantage: we couple ourselves to markdown-it's internal design more)?

lib/gfm/override-image-renderer.js
```diff
@@ -0,0 +1,23 @@
+var imageRule = require('../rules/inline_rules/image')
+
+module.exports = function (md, options) {
+  // Unfortunately, there's no public API for getting access to the existing
+  // installed parsing rules; rather than import the 'reference' rule directly
+  // from markdown-it just so we can re-install it into the parser with the
+  // 'alt' chain set up correctly, here we're just using internal utility
+  // methods to modify it in place at runtime.
```
revin: Looks like this comment didn't get updated when you created this file?

sjking: You're correct, I missed that. Updated now by removing unrelated comments.

sjking:
This PR overrides the default image rule from `markdown-it/rules/inline_rules/image`. Because this rules depends on internal helper modules, I also copied those over. The only real change to how those modules work is in the parsing of link destinations: https://github.com/sjking/marky-markdown/blob/6f2b7c4433268f013d92ad0f64e409988b823382/lib/rules/helpers/parse_link_destination.js#L22. Some other code looks *slightly* different due to style/linting differences between the two projects.

revin:
Thanks for cleaning up that file. 😄 It might take a little time to discuss what we want the strategy to be with regard to the question of importing vs copying the helper modules; I was going to suggest possibly redoing the file organization a bit rather than being an exact copy of how markdown-it does it (like, probably move the actual rule body into lib/gfm/override-image-renderer.js, put the other stuff currently nested inside lib/rules/ into a flat directory lib/util/ or something, perhaps; stuff like that), but it's probably too early for that, pending the result of the discussion.

It's kind of a bummer how much code we have to copy just for this kind of change (nothing against what you've done here @sjking; this is great; it's just the markdown-it design); I wonder if it's worth filing an issue on markdown-it to see if they're willing to make that part pluggable.

revin:
ok here it is: https://github.com/markdown-it/markdown-it/issues/299. I'm not super hopeful about this since there isn't a ton of incentive on their end to overhaul the internals just for us, but you never know.

revin:
Hey @sjking it looks like markdown-it 8.1 will solve the issue of making the helper functions pluggable, thanks to [f5a06ec0](https://github.com/markdown-it/markdown-it/commit/f5a06ec0b626d857d06382e7b8709d943d6f2000). 🎉 🎉 🎉 

That means in the plugin here we can override `md.helpers.parse_link_destination`, `.parse_link_label`, and `.parse_link_title` and skip having to install/replace any parser rules. 😄 

Also, TIL that `md.utils` has a reference to the markdown-it utilities module, so being able to use that should allow us to skip needing to import that file, I think. Mind trying that out? In #283 I was able to call, e.g., `state.md.utils.isSpace(...)` and take out the inlined version I'd created.

sjking:
@revin I was able to actually modify the `state` object properties for `parse_link_destination` with a customized function. However that appears to be a shared object so it might cause side-effects elsewhere. For example, the `parse_link_destination` rule is shared by the rule that parses anchor , or `href` elements. Modifying it for the `image` rule will also effect how the `href` rule works.

revin:
@sjking oh good call. Hmm, maybe we do need a parsing rule override that just temporarily replaces the helper methods and calls the original rule, then restores the helpers back.

GH markdown doesn't allow spaces in normal link hrefs? I can't remember off the top of my head.

revin:
After thinking about this some more, I think space-supporting method overrides probably are indeed the desirable thing for us. If we strip this PR down to just those, as well as add tests that also deal with spaces in normal links, does the rest of the test suite still pass?

sjking:
@revin That sounds reasonable. I'll have to write tests to cover the cases that are not for images.

sjking:
I think this should work for parsing link destinations and allowing spaces in path names. I added some new tests as well.

revin:
Oops it looks like other commits created a merge conflict for this @sjking 😬; would you mind rebasing it onto the current master, please? Thanks!

sjking:
Is this sufficient?

ashleygwilliams:
LGTM- @revin?

revin:
@ashleygwilliams GH is showing a bunch of unrelated commits here, is that just an artifact of working across forks, or is that going to end up duplicated those in the history?

ashleygwilliams:
@revin- oh good catch. i don't think they will be duplicated but one of us could manually merge this to ensure that- lemme know if you'd like me to!

revin:
@ashleygwilliams if you have time now, go for it. I can look at it later today, but not for a few hours at least.

revin:
Tested merging it manually locally, and yes it ends up creating two separate branches that get merged back together and back into master. I think that's probably because we rebased/squashed stuff that got committed to master in between this one being branched off and finished. The end result in master is correct; the history is a little confusing, is all.

I tried rebasing the branch onto master (which worked OK, though I had to resolve a few merge conflicts along the way) and then squashing the whole thing down into a single commit, and in the end it ends up with a pretty tidy change set that modifies lib/render.js & test/sanitize.js, and adds lib/gfm/override-link-destination-parser.js. That looks spot on. @sjking do you want to try and replicate that work on your side and then force push to this branch? I can certainly help if what I'm describing doesn't make sense. Or I can submit the test branch I now have here locally if you'd rather me do that. Up to you!

sjking:
@revin OK sure, I didn't know that the goal was to rewrite the commit history in that way. That would effectively turn this pull request into a single commit that adds `lib/gfm/override-link-destination-parser.js` and modifies `lib/render.js` and `test/sanitize.js`, if I'm not mistaken. If you don't mind waiting till Monday then I'll have some time to make those changes.

revin:
@sjking awesome! Please take your time; no rush at all. (@ashleygwilliams please also feel free to overrule my rebase/squash opinion if you want)

ashleygwilliams:
@revin your plan sounds good to me :) -- thanks again @sjking !

revin:
Hi @sjking just checking in. I wasn't able to do much marky work during the holiday season, so now that I'm more around I thought I'd look at the status of our PRs. No pressure or hurry or anything; I was just wondering what your plans were for this. Are you still working on it at all? If you've lost interest or hit a roadblock or whatever, I'm happy to help push it through, or take it over altogether if you don't have the time, or whatever you like really. Just let me know!

Thanks so much for what you've done so far on this; it's super cool ✨🎈🐙 🎈 ✨

sjking:
Hi @revin thanks for checking in. I had done some work on it about a month ago, but there were some tests failing at that time. Everything looks fine now that I rebased it on master.

revin:
@sjking YESSSSSS thanks so much; this is nice and clean. Mergin'

PR #281, opened 2016-10-27
GitHub list parsing

jamestautges:
This isn't complete by any means, but it does implement the basics of turning indentation into list levels. I copied the [implementation github uses](https://github.com/vmg/redcarpet/blob/master/ext/redcarpet/markdown.c) as much as I could. There are certain things that markdown-it doesn't like you doing--such as calling the inline parser from a block rule--that are essential to this working properly. It's definitely possible, but it'll take another day or two of jiggery-pokery :sparkles: :tophat: :rabbit: :sparkles:.

There are going to be tons of edge cases, so I'd like some help trying to identify them for unit tests.

### DIFF
--- a/lib/plugin/list.js
+++ b/lib/plugin/list.js
@@ -55,6 +55,7 @@ module.exports = function (md, opts) {
 
 	    if (line.trim() === '') {
 		inEmpty = 1
+		lineIndex++
 		continue
 	    }
 	    
--- a/lib/plugin/list.js
+++ b/lib/plugin/list.js
@@ -7,6 +7,20 @@ var isOrderedItem = function (line) {
     var match = line.match(/^( *)([1-9][0-9]*)\. /)
     return match
 }
+
+var hRule = function (line) {
+    var stripped = line.trim()
+    var c = stripped[0]
+    if (c !== '-' && c !== '*' && c !== '+') return false
+
+    var n = 0
+    for (var i = 0; i < stripped.length; i++) {
+	if (stripped[i] === c) n++
+	else if (stripped[i] !== ' ') return false
+    }
+
+    return n >= 3
+}
   
 module.exports = function (md, opts) {
     var listItemParser = function (state, startLine, endLine, silent) {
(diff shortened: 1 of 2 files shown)


revin:
Oh this is awesome! Thank you!

Yes, with regard to edge cases, I think we're going to need to create some kind of list parsing torture test in test/fixtures/, and then fill it up with all kinds of malformed markdown lists 😬

It'll be amazing 😛

ashleygwilliams:
hey @jamestautges ! checking back in - do you need any help working on this? we'd love to see it complete and are happy to help! let us know :)

jamestautges:
Sorry everyone. I hit an unexpectedly busy couple of days, but I should have more time this weekend. I'd like to get both of these PRs done before Monday.

ashleygwilliams:
no worries! was just making sure that you had what you needed, no rush :)

jamestautges:
Alright, that should do it. It ended up being much easier than I thought once I dug into markdown-it's API. This commit includes changes to how code fences are handled in lists, so I think that might address the original issue, but I didn't test it so I'm not confident. Unit test time I guess.

jamestautges:
Oops, broke task lists :grimacing: . I'll fix it tomorrow.

lib/plugin/list.js
```diff
@@ -0,0 +1,265 @@
...
+
+var isOrderedItem = function (line) {
+  var match = line.match(/^( *)([1-9][0-9]*)\. /)
+  return match
+}
+
+var prefixCodefence = function (line) {
+  if (line.length < 3) return 0
+
+    // This is how it looks in redcarpet and I didn't feel like making it better
```
jamestautges: The weirdest part is that they use ```while (i < 4 && line[i] === ' ') i++``` just 40 lines later. But hey, it beats the bit-flags they pass through every parsing function.

revin:
Hi @jamestautges just checking in now that the holiday season is mostly over. No pressure at all or anything, but I was wondering if you were still planning on trying to get the tests figured out? If you've lost interest and/or moved on that's totally cool; I'm excited about this change though 🎉🎉🎉 so I'm happy to take it over if you like 😄

PR #282, opened 2016-10-31
separate packagize code into separate dep, pkgize

ashleygwilliams:
pulls packagize code into it's own pkg, `pkgize`: https://www.npmjs.com/package/pkgize

had to pull `token-util` into `pkgize`... should we make that it's own separate pkg? i dont like the repeated code but also don't know if we'll even save pkgize so maybe it doesn't matter?


revin:
@ashleygwilliams Can you please remind me of the reasoning behind this, again? Is it to keep the "npm-ish" stuff out of marky so marky can stick to being a GitHub-style markdown parser? If so, do you think [cdn.js](https://github.com/npm/marky-markdown/blob/30865e4c37fae402d9e18b83a669bf8103259d27/lib/plugin/cdn.js) should be extracted as well?

revin:
oh, also re: `token-util`, it turns out the one function `pkgize` needs is `getText`, and no other code in marky needs it, so you should be able to just move `getText` all by itself without issue.

ashleygwilliams:
@revin - exactly. i'd like marky to be just a GFM parser, and right now it's doing some other stuff. in the long run- i'm not sure the website wants it to do this stuff and so in preparation, i'm separating it out. i think ideally it's code that we can just get rid of all together.

ashleygwilliams:
this is waiting on me moving over the tests. also needs a rebase as this point.

PR #283, opened 2016-11-01
Stop escaping characters inside code blocks inside table cells

### LINKED ISSUE #252: single backslash in an inline code block inside table parsed differently than GitHub MD
from https://github.com/npm/www/issues/31 via @poiuiopoiu 

NPM appears to use a parser that follows the common mark spec and it differs from github in a few ways.  My issue is writing a single backslash in an inline code block.

The Github way is 

```
`\`
```

But that trips up the NPM parser, since common mark treats that as an escaped backtick.  The right way to do it for NPM is

```
`\\`
```

But that messes up github, which interprets it as 2 backslashes.

I'd guess that most modules use the github markdown style, so I guess I would consider this a npm website bug

revin:
Here's a PR that addresses the difference in opinion between GitHub and CommonMark with regard to escaping characters inside code blocks inside table cells.

Most of it is just importing the stock markdown-it table parser and modifying it slightly (mostly just adding `&& !backticked` on what is now line 43 of the `table.js` here, but also turning the file into a markdown-it plugin we can use in `render.js`). Also added a test fixture based on the `badness-10000` package, with a couple of additions to make sure we're covering our bases.

Fixes #252


revin:
I tagged this with `semver-major` because technically our generated HTML is different after this, but if anyone thinks we can consider it a bug fix and put it in a patch release, I won't argue with that either.

ashleygwilliams:
our next goal is 10.0.0 so i'm cool with this being in that version. since we've only gotten the one complaint i dont think we need to rush this getting out. if we hear more reports between now and 10.0.0 let's consider a patch., but otherwise plan for 10.0.0.

ashleygwilliams:
hey @revin you mentioned this in #280... does that mean you want to change how this works or..? i think it is otherwise good to go. lemme know!

revin:
@ashleygwilliams already did! This PR was easy to update because there was just the one function invocation to change, so it's good to go for now, as far as I can see.

ashleygwilliams:
ok this looks good to me! mergin'

Issue #247, opened 2016-09-23
Allow additional markdown-it plugins


sqrrrl:
Any thought to allowing the registration of additional markdown-it plugins? Specifically, I'd like to add support for attributes, but there doesn't seem to be a good way to register additional plugins. Would be nice to be able to pass in additional plugins when parsing, like:

``` javascript
marky('blah blah', { plugins: [ require('markdown-it-attrs')] })
```

If open to it, can submit a pull request.

revin:
Hi @sqrrrl thanks for filing!

On one hand I do like the idea, but on the other hand it feels sort of weird to be exposing markdown-it, which to date has been treated as an internal implementation detail. Adding this capability would tie our version numbering to markdown-it's versioning a bit too, since then we have to track plugin compatibility when markdown-it updates. _Plus_, it's not hard to argue that adding plugins is outside the scope of what marky's trying to do, which is to replicate GitHub's markdown processing (so if there's a plugin that gets us closer to GH rendering, we should just adopt it).

*However*, It's seems pretty useful though, like I could even imagine using it during testing/development of new marky features.

@ashleygwilliams what are your thoughts on adding "secret"/likely-to-break APIs, and explicitly opting to never worry about backwards compatibility?

jdarling:
I just ran into the need to add custom tags to my document, it would be great to either support existing markdown-it plugins or some type of marky plugin for extension.  Personally I think a custom marky style plugin would be a better choice, nothing that says someone couldn't then create a plugin to "port" mardown-it style plugins to marky.

revin:
@sqrrrl we just merged #285, which should allow you do do what you need (see a code example in the PR description). It won't be available in the public package until the next release, but if you clone from master you can check it out.

I'm going to close this now, but if this isn't sufficient for whatever reason, please feel free to reopen.

revin:
oh hmm I never got a notification for your comment @jdarling, sorry about that. I'm not _completely_ clear on what you want; does the `getParser` API give you what you need? If not, maybe open a new issue covering it? Thanks!

PR #285, opened 2016-11-03
Add new getParser public API

revin:
Here's a speculative change that adds a new public API:

### getParser([options])

Returns a marky-markdown-flavored markdown-it parser object. You can attach other markdown-it plugins, inspect the internals for debugging, etc... For example:

```js
var marky = require('marky-markdown')
var someMarkdownItPlugin = require('some-fake-markdown-it-plugin')

var parser = marky.getParser({ package: somePackage, highlightSyntax: false })

parser.use(someMarkdownItPlugin)

var html = parser.render(html)
```

`marky.getParser(options).render(markdown)` is the exact same thing as the existing `marky.render(markdown, options)`

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -46,7 +46,11 @@ if (typeof process.browser === 'undefined') {
   cleanup(highlighter.registry.grammars)
 }
 
-var render = module.exports = function (html, options) {
+var render = module.exports = function (markdown, options) {
+  return render.getParser(options).render(markdown)
+}
+
+render.getParser = function (options) {
   var mdOptions = {
     html: true,
     langPrefix: 'highlight ',
@@ -79,7 +83,11 @@ var render = module.exports = function (html, options) {
   if (options.highlightSyntax) parser.use(codeWrap)
   if (options.serveImagesWithCDN) parser.use(cdnImages, {package: options.package})
 
-  return githubLinkify(parser).render(html)
+  return githubLinkify(parser)
+}
+
+render.renderPackageDescription = function (description) {
+  return MD({html: true}).renderInline(description)
 }
 
 var mappings = {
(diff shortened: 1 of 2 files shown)


revin:
The main reason I'm proposing this is that I'm working on a live in-browser marky testing app, and having access to the parser made it possible to get access to the markdown-it token stream so I could do a Debug view. It also ends up addressing #247, which is nice.

If everybody likes this change, I can update the documentation and add some tests.

lib/render.js
```diff
@@ -79,7 +83,11 @@ var render = module.exports = function (html, options) {
   if (options.highlightSyntax) parser.use(codeWrap)
   if (options.serveImagesWithCDN) parser.use(cdnImages, {package: options.package})
 
-  return githubLinkify(parser).render(html)
+  return githubLinkify(parser)
+}
+
+render.renderPackageDescription = function (description) {
```
ashleygwilliams: what does `PackageDescription` mean here? feel like we may soon be removing this when we get rid of `packagize.js` with #282 

otherwise this is a misleading name. shouldn't affect this PR, but we should change that if it means the full `README`

revin: That's the name this method has had since before I started working on marky; last I saw many months ago, the npm web site was using it for rendering package.json `description` fields into HTML for use in on package pages.

If they're not using it anymore, I'd be in favor of deprecating/removing it. People will be able to do basically the same thing by calling `marky.getParser().renderInline(markdownSnippet)`, but honestly it wouldn't surprise me if nobody at all is using it anyway.

revin: oh yeah, and this has absolutely nothing to do with the packagize stuff!

and finally, the only reason it shows up in the diff here is because I moved it up earlier in the file so it grouped the module exports together 😕 

yay legacy

ashleygwilliams:
haha ok cool so- i think this looks awesome and i'm really excited about it. gonna add getting rid of that package description thing to #282  since we (cc/ @jefflembeck  @aredridel ) are planning on getting rid of that pesky readme + desc weirdness anyways

omg 10.0.0 is gonna be so good

ashleygwilliams:
ok! travis is FINALLY green lol. mergin'

Issue #286, opened 2016-11-04
Need docs & tests for `getParser`, also sanitization support


revin:
I forgot to add the tests and doc changes to #285, so we _have_ a `getParser` API, but there's nothing in the README and no test coverage.

Also, it turns out the initial implementation doesn't properly honor the `{sanitize: true}` option. Need to fix that.

ashleygwilliams:
closed by #287

PR #287, opened 2016-11-04
Add tests and documentation for getParser

revin:
Here are docs and tests for #286. This also updates the parser object returned by `getParser` such that it will sanitize the generated HTML when `{sanitize: true}` is passed (or not specified; it's the default).

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,25 @@ The default options are as follows:
 }
 ```
 
+### Low Level Parser Access
+
+If you need lower level access to the markdown-it parser (to add your own
+[markdown-it plugins](https://www.npmjs.com/search?q=markdown-it-plugin), for
+example), you can call the `getParser` method:
+
+```js
+var parser = marky.getParser()
+parser.use(someMarkdownItPlugin)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -77,6 +77,25 @@ The default options are as follows:
...
+
+```js
+var parser = marky.getParser()
+parser.use(someMarkdownItPlugin)
+var html = parser.render("# markdown string")
+```
+
+`getParser` takes an optional `options` argument, the same format as the main
+marky-markdown export function. If you omit it, it uses the same default options
+describe above.
```
ashleygwilliams: maybe add a link to that default options part of the code?

revin: they're listed directly above this section in the readme already, do you want me to make that a link too?

heh if I change this I'm totally making a unit test that makes sure those links don't get stale if we change index.js 😛

ashleygwilliams: nope i'm just a super lazy reviewer. don't worry about! should be fiiiine :)

ashleygwilliams:
ok let's merge this!

PR #289, opened 2016-11-20
use `anchor` as class name instead of `deep-link`

### LINKED ISSUE #288: use `anchor` instead of `deep-link` as class for header links
GH uses `anchor`. in an attempt to match them perfect, we should change this. this is a breaking change.

ashleygwilliams:
fixes #288


ashleygwilliams:
thoughts @revin?

revin:
oh nice I noticed this the other day, thanks for fixing it 👍

PR #292, opened 2016-11-20
use octicon octicon-link instead of deep-link-icon

### LINKED ISSUE #291: use `octicon octicon-link` instead of `deep-link-icon`
because GH

ashleygwilliams:
fixes #291


revin:
✨🐬✨

PR #293, opened 2016-11-20
add `aria-hidden=true` to adding headings

### LINKED ISSUE #290: add `aria-hidden=true` to deeplinks
GitHub does this, so we should do this.

(GHDTSWSDT)

ok yeah that acronym isn't great. lol.

ashleygwilliams:
fixes #290

### DIFF
--- a/lib/plugin/heading-links.js
+++ b/lib/plugin/heading-links.js
@@ -49,6 +49,7 @@ var headings = module.exports = function (md, options) {
       // add some link attributes
       linkOpen.attrSet('id', headings.prefix + postfix)
       linkOpen.attrSet('class', headings.className)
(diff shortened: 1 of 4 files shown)


ashleygwilliams:
tests are not passing and i'm not sure why right now, ~work in progress~

ashleygwilliams:
had to fix sanitize. this is read for review now

revin:
Nice! Although #289 changes `deep-link` to `anchor` so I guess we need to change these tests for that now since I just merged the other one.

ashleygwilliams:
@revin yeah i can fix that

ashleygwilliams:
OK! @revin this should be good now

PR #299, opened 2016-11-20
allow style attr on img tag

### LINKED ISSUE #298: do not strip `style` attr from `img` tag
GHDSWS

(github doesn't so we shouldn't)

ashleygwilliams:
fixes #298


revin:
aw yeah you're the best 🎷🐡

PR #307, opened 2016-11-22
Zip together markdown and HTML heading tests

### LINKED ISSUE #301: what is the difference between `test/headings` and `test/html-headings`
it's not clear to me and i def wrote some redundant tests in my recent PRs, afaik

thoughts? i'd like to see this clarified or combined into a singular test file

revin:
Fixes #301


ashleygwilliams:
really love this! thank you

PR #308, opened 2016-11-23
Replace github paths in `html_inline` links/images

### LINKED ISSUE #304: need inline-html rule for the `github` plugin
currently, `github` doesn't work for img/links written in inline html

revin:
Fixes #304

### DIFF
--- a/lib/plugin/github.js
+++ b/lib/plugin/github.js
@@ -5,6 +5,43 @@ var gh = require('github-url-to-object')
 var URL = require('url')
 var path = require('path')
 
+function buildImageUrl (repository, url) {
+  var prefix = 'https://raw.githubusercontent.com/'
+  return prefix + path.join(repository.user, repository.repo, 'master', url.href)
+}
+
+function buildLinkUrl (repository, url) {
+  return repository.https_url + path.join('/blob/master/', url.href)
+}
+
+// search the provided HTML snippet and rewrite the attribute on the given tag
+// using the repository object provided
+function replaceTagAttribute (html, tag, attribute, buildUrl, repository) {
+  // look for the attribute's value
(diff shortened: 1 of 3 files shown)


lib/plugin/github.js
```diff
@@ -54,4 +82,23 @@ module.exports = function (md, opts) {
...
     }
   }
+
+  // rewrite image locations in inline HTML to be fully qualified github URLs
+  var originalHtmlInlineRule = md.renderer.rules.html_inline
+  md.renderer.rules.html_inline = function (tokens, idx, options, env, self) {
+    var content = tokens[idx].content
+    var repoPath = path.join(repo.user, repo.repo, 'master')
+    var imageBaseUrl = 'https://raw.githubusercontent.com/' + repoPath
+    var newHtml
```
revin: I'm not super happy with the nested `if()` structure here, but just doing, e.g.,
```js
newHtml = replaceTagAttribute(content, 'img', 'src', imageBaseUrl + '/') ||
          replaceTagAttribute(content, 'a', 'href', repo.https_url + '/blob/master/')
```
looked bad when the `standard` formatting was applied (leading whitespace on the second line gets removed)

ashleygwilliams:
assuming that's the case (which if it weren't would mean that we had bigger problems to begin with)- this looks great for merge 🐬 ✨

revin:
Yes it's only for GH repos thanks to `github-url-to-object` doing the URL parsing (see [github.js#L50-52](https://github.com/npm/marky-markdown/blob/2dbdca0579038b50bc7ae4764beea36f5e972a8f/lib/plugin/github.js#L50-L52)); it returns `null` for non-GH URLs

ashleygwilliams:
a+. mergin!

PR #313, opened 2016-11-24
Don't needlessly escape characters in regular expressions

### LINKED ISSUE #312: Unnecessarily escaped characters in regular expressions break our build
Per #311, the eslint `no-useless-escape` rule stops our build from completing due to a few places where we're escaping characters unnecessarily.

revin:
Fixes #312


revin:
Merging this because it's causing greenkeeper to fail on everything

PR #316, opened 2016-12-01
Update to markdown-it 8.2 and remove our table escaping rule

revin:
The resolution of [markdown-it/markdown-it@303](https://github.com/markdown-it/markdown-it/issues/303) makes our table code escaping rule obsolete, so here's a PR to update to markdown-it 8.2 and remove the unnecessary code


ashleygwilliams:
this is pretty awesome.

PR #322, opened 2017-01-03
Add html_block github image/link replacer

### LINKED ISSUE #320: package images failing to load
We got a report that this logo image on the `kaiser-crawler` package page is not loading: 
https://www.npmjs.com/package/kaiser-crawler

screen shot 2017-01-03 at 11 49 43 am

@ashleygwilliams suggested this was a bug in marky and could (probably) be fixed shortly, thanks for looking!

revin:
Fixes #320 by also scanning `html_block` tokens for links & images to rewrite with GitHub URLs

### DIFF
--- a/lib/plugin/github.js
+++ b/lib/plugin/github.js
@@ -18,7 +18,7 @@ function buildLinkUrl (repository, url) {
 // using the repository object provided
 function replaceTagAttribute (html, tag, attribute, buildUrl, repository) {
   // look for the attribute's value
-  var regex = new RegExp('^<\\s*' + tag + '[^>]*\\b' + attribute + '\\s*=\\s*', 'i')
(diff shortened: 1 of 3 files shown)


lib/plugin/github.js
```diff
@@ -18,7 +18,7 @@ function buildLinkUrl (repository, url) {
 // using the repository object provided
 function replaceTagAttribute (html, tag, attribute, buildUrl, repository) {
   // look for the attribute's value
-  var regex = new RegExp('^<\\s*' + tag + '[^>]*\\b' + attribute + '\\s*=\\s*', 'i')
+  var regex = new RegExp('<\\s*' + tag + '[^>]*\\b' + attribute + '\\s*=\\s*', 'i')
```
revin: This regex was originally anchored at `^` because in `html_inline` tokens, the inline tags always start at the beginning of the token content. But with `html_block` we might have a whole set of complex markup, so we can't do that for those.

PR #323, opened 2017-01-03
Enforce that if options are passed, they must be an object

### LINKED ISSUE #321: assert that options is an object
I ran into an issue recently where I was accidentally passing in a string for options rather than an object; rather than exploding, this resulted in an instance of marky that seemed to have no defaults enabled.

I think better behavior would be an assert.

revin:
Fixes #321.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 var defaults = require('lodash.defaults')
+var isPlainObj = require('is-plain-obj')
 var render = require('./lib/render')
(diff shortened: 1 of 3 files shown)


revin:
Here's a browserify friendly non-`assert` way of handling the `options` object, which may or may not be acceptable given our overall goals.

ashleygwilliams:
this is awesome woo

PR #325, opened 2017-01-11
Non Greedy HTML Block Parsing

### LINKED ISSUE #317: Continues to parse markdown as inline HTML when there is no blank line between them
Cross-post from @avrahams  https://github.com/npm/www/issues/61

The **_bold & italic_** notation is not shown correctly in npm 
https://help.github.com/articles/basic-writing-and-formatting-syntax/#styling-text

See [IoT virtual device module page](https://www.npmjs.com/package/node-red-contrib-iot-virtual-device)

in npm
image

in github
image

revin:
Here's a markdown-it plugin and tests to cover what's in #317, where we have markdown appearing after an HTML block without the CommonMark-mandated blank line in between.

This is fairly tricky, because the reasoning behind CommonMark's behavior here is completely sensible: once an HTML block is started, how can we tell it's actually finished? What this PR does is check to see if the lines in the parsed HTML block match any of the higher priority "block" type markdown parsing rules (list, heading, etc...), and if so, adjusts the parsing state to cut the HTML block off at the first sight of valid markdown.

Note that this only works its magic on "block" type rules. Before this PR, the following gets parsed as one big `html_block` token:
```markdown
<img src="path/to/image.png"/>
- item 1
- item 2
```
and now that's an `html_block` followed by an unordered list. However:
```markdown
<img src="path/to/image.png"/>
some +other+ **formatted* _text_
```
will still be a single `html_block`.

Fixes #317

### DIFF
--- a/lib/gfm/abstemious-html-block.js
+++ b/lib/gfm/abstemious-html-block.js
@@ -1,0 +1,90 @@
+module.exports = function (md, options) {
+  // Unfortunately, there's no public API for getting access to the existing
+  // installed parsing rules; rather than import the 'html_block' rule directly
+  // from markdown-it just so we can wrap it with our custom logic here, we just
+  // use internal utility methods to modify it in place at runtime.
+  //
+  // Normally, CommonMark HTML blocks are terminated by a blank line. However,
+  // GitHub will terminate blocks early if the parser encounters Markdown in
+  // subsequent lines even before it reaches a blank one. To match that
+  // behavior, what we do here is:
+  //
+  //   1 Run the html_block rule to see if the current line is the start of an
+  //     HTML block
+  //
+  //   2 If so, walk through the block line by line, running each line through
+  //     each of the higher priority block rules (in validation mode, since
+  //     we're only testing for matches) until we find either a blank line or a
+  //     line that validates as one of those rules (e.g., a list, heading, etc...)
+  //
+  //   3 If we found a blank line or a markdown line, update the bounds and
+  //     content of the html_block that was created in step 1 and reset
+  //     state.line such that markdown-it will resume parsing from that point
+  //
+  // That means that this (among other things) works:
+  //
+  //   <img src="some/image.png"/>
+  //   - the start of a list
+  //   - the second item
+  //   - etc
+  //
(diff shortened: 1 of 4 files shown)


lib/gfm/abstemious-html-block.js
```diff
@@ -0,0 +1,90 @@
+module.exports = function (md, options) {
+  // Unfortunately, there's no public API for getting access to the existing
```
ashleygwilliams: you are an 👼 for writing this epic comment. thank u

lib/render.js
```diff
@@ -19,6 +19,7 @@ var htmlHeading = require('./plugin/html-heading')
 var relaxedLinkRefs = require('./gfm/relaxed-link-reference')
 var githubHeadings = require('./gfm/indented-headings')
 var overrideLinkDestinationParser = require('./gfm/override-link-destination-parser')
+var abstemiousHtmlBlocks = require('./gfm/abstemious-html-block')
```
ashleygwilliams: lolmg gonna need a dictionary to read our source now haha

ashleygwilliams:
p.s. omg ugh https://johnmacfarlane.net/babelmark2/?normalize=1&text=%3Cimg+src%3D%22some%2Fimage.png%22%2F%3E%0A+%2B+the+start+of+a+list%0A+%2B+the+second+item%0A+%2B+etc

ashleygwilliams:
any idea what the performance hit (if any) this is, @revin? i'm not sure how common it is in the registry in general so it may not matter, but i'm curious

revin:
I didn't run any bulk numbers, but I certainly can. For documents that don't contain HTML blocks, it's no slower, of course, but then the cleaner a particular HTML block is (i.e., the longer it goes without containing interrupting markdown), the worse the slowdown. The pathological case would be to feed marky a large HTML document, and the rules would get run each time for every line, _after_ the document was parsed and turned into a giant `html_block` token in the first place 😢

ashleygwilliams:
well luckily, now that npm pre-renders readmes, the performance issue is not a SUPER big deal. i'm down to merge this! one day we should have a hack day where we build some benchmarking tools for realz

PR #329, opened 2017-01-18
Allow whitespace between image/link labels and paths

### LINKED ISSUE #328: Allow whitespace inside links and images
Hello, I noticed that some shields from http://shields.io/ do not work. It displays like this: 

screen shot 2017-01-15 at 2 17 25 pm

I have tried multiple things, such as editing spacing, but it doesn't work. The shields display correctly on Github. Also, I looked at other issues, which said it was because the description was truncated, but I do not think this was truncated.

Thank you.
Andrew

revin:
It's kind of a bummer, but the type of looser parsing github does on links and images requires us to completely overwrite the markdown-it image and link rules. So, here it is.

Fixes #328

### DIFF
--- a/lib/gfm/image.js
+++ b/lib/gfm/image.js
@@ -1,0 +1,179 @@
+// CommonMark specifies that images have no whitespace between the alt text and
+// src, e.g.:
+//
+//     ![alt text](path/to/the/image)
+//
+// However, GitHub allows you to put whitespace between the `]` and `(`
+// characters, like so:
+//
+//     ![alt text]    (path/to/the/image)
+//
+// To account for the difference, we need to override markdown-it's image rule.
+// This file is a copy of the original markdown-it source, plus the necessary
+// modifications, wrapped as a markdown-it plugin.
+
+// Process ![image](<src> "title")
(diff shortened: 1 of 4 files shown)


revin:
Sure, I can take a look at doing that. I thought about it while I was working on this, and decided to be conservative and do as little modification to the original source as possible. But I think at least some of these loops that walk through the markdown can be refactored easily.

ashleygwilliams:
<3<3<3 thank u. u r amazing. ur work is SUPER APPRECIATED.

ashleygwilliams:
it occurs to me that if you want to just ships this as is we could add an issue for refactoring it later. and then that way we could 10.0 soon

ashleygwilliams:
also i merged this out of order so i caused a conflict i'm sorrrrrrrrrrry 😩

revin:
>it occurs to me that if you want to just ships this as is we could add an issue for refactoring it later. and then that way we could 10.0 soon

That's a good idea. I'd hate to hold everything up for what would be a patch-release-type change to the code. So I'd vote for that.

PR #342, opened 2017-02-23
feat: deprecate 0.12 as it is no longer a maintained version of node.js

ashleygwilliams:


ashleygwilliams:
i'm rushing this through rn @revin to speed up travis. we already chatted about it so i hope it is still oK!

revin:
absolutely, thanks for being on top of it 👏🏽

PR #346, opened 2017-02-28
Add comment with marky version and issues URL to rendered HTML in debug mode

### LINKED ISSUE #344: add option that would include a debug message in generated HTML
came from a convo with @chrisdickinson and @bcoe 

<hr/>

if a debug option is set to true, add an HTML comment that looks something like this to the top of the generated HTML:

```html
<!-- this HTML was generated using marky-markdown version ##.##.##. see an issue? file at <issue link>. please include the version in your issue. thanks for using marky! -->
```

revin:
Fixes #344.

This does result in the contents of our package.json being included in the browserify build. Do we care about that at all?

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@ var defaults = require('lodash.defaults')
 var isPlainObj = require('is-plain-obj')
 var render = require('./lib/render')
 var sanitize = require('./lib/sanitize')
+var markyInfo = require('./marky.json')
 
 var defaultOptions = {
   sanitize: true,
@@ -41,6 +42,18 @@ var marky = module.exports = function (markdown, options) {
   if (options.sanitize) {
     log('Sanitize malicious or malformed HTML')
     html = sanitize(html)
+  }
+
+  if (options.debug) {
+    var debugHeader =
+      '<!--' +
+      ' this HTML was generated using marky-markdown version ' + markyInfo.version + '.' +
+      ' see an issue? file at ' + markyInfo.issuesUrl + '.' +
+      ' please include the version in your issue. thanks for using marky!' +
+      ' to learn more, visit ' + markyInfo.repositoryUrl + '.' +
+      '  -->'
+
+    html = debugHeader + '\n' + html
   }
 
   return html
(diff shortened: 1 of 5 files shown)


ashleygwilliams:
is this still a WIP @revin ?

index.js
```diff
@@ -43,6 +44,18 @@ var marky = module.exports = function (markdown, options) {
     html = sanitize(html)
   }
 
+  if (options.debug) {
+    var issueUrl = markyPackage.repository.url + '/issues'
+    var debugHeader =
+      '<!--' +
+      ' this HTML was generated using marky-markdown version ' + markyPackage.version + '.' +
+      ' see an issue? file at ' + issueUrl + '.' +
+      ' please include the version in your issue. thanks for using marky!' +
```
ashleygwilliams: maybe add a "to learn more about marky, visit {github URL}"

ashleygwilliams:
in general i think this looks good! might just add that one thing but otherwise 🐑 ⛵️ 🚢 :shipit: 🚀

revin:
@ashleygwilliams Yeah I re-WIP'd it because I'd remarked about the package.json being included in the browserify bundle, and @bcoe suggested maybe doing the version in some kind of build step, so I was thinking about doing that instead. Am absolutely open to opinions.

revin:
@ashleygwilliams ok it's been unWIP'd; I changed it so now we do a prebuild script that generates the necessary info in a new file called `marky.json`. And then marky pulls that in rather than the entire `package.json`. Feels like a better solution than what I had before, IMHO.

package.json
```diff
@@ -4,6 +4,7 @@
   "description": "npm's markdown parser",
   "main": "index.js",
   "scripts": {
+    "prebuild": "node prebuild.js",
```
bcoe: clever, I like the simplicity of this approach; and the fact that you don't need to remember to update variables on each release.

revin: Hmm, speaking of which, I suppose it would be helpful to rebuild the `marky.json` in a `postversion` hook too. Lemme do that.

revin:
All righty I also set up a `postversion` hook so it automatically rebuilds all of this (including the browser build in `dist/`) when we do `npm version [whatever]`

Issue #350, opened 2017-03-23
GitHub URL conversion fails if repo has no `master`


zkat:
Additionally, it will invariably point to `master` even if the repo has been configured to use a different branch as its default.

I'm posting this issue to propose a PR where `master` gets replaced with `HEAD` (which will point to the default remote branch). Essentially, this would make it so that `[readme](README.md)` would always go to the same URL people would end up in if they visited the repository directly at `https://github.com/user/proj`.

PR #351, opened 2017-03-23
fix(github): use default branch instead of hardcoding `master`

zkat:
Fixes: #350


revin:
Hey @zkat this is a h*ckin great idea, thanks! mergin'

PR #357, opened 2017-04-12
fix(sanitize): prefix ids from user generated html

### LINKED ISSUE #356: treat ids in html as we would other user generated ids
e.g.

```
<div id="add-billing-info">Foo.</div>
```

generates 
```
<div id="add-billing-info">Foo.</div>
```

when it should be

```
<div id="user-content-add-billing-info">Foo.</div>
```

ashleygwilliams:
fixes #356 
this will be a 9.0.3 release that we should do before tomorrow

### DIFF
--- a/lib/sanitize.js
+++ b/lib/sanitize.js
@@ -131,6 +131,6 @@ function prefixHTMLids (tagName, attribs) {
   return {tagName, attribs}
 }
 
-function isAlreadyPrefixed(id, prefix) {
+function isAlreadyPrefixed (id, prefix) {
   return id.includes(prefix) && id.length > prefix.length
 }
--- a/lib/sanitize.js
+++ b/lib/sanitize.js
@@ -1,6 +1,18 @@
 var sanitizeHtml = require('sanitize-html')
-var sanitizer = module.exports = function (html) {
-  return sanitizeHtml(html, sanitizer.config)
+var sanitizer = module.exports = function (html, options) {
+  var config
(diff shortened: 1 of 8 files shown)


lib/sanitize.js
```diff
@@ -111,3 +123,12 @@ function sanitizeCellStyle (tagName, attribs) {
   var result = alignmentRegEx.exec(attribs.style || '')
   return result ? cell(result[1]) : cell()
 }
+
+function prefixHTMLids (tagName, attribs) {
+  if (attribs.id && attribs.id.replace('user-content-', '').length > 0) {
```
ChALkeR: What is the second check supposed to do here? It looks equivalent to `attribs.id !== 'user-content-'` atm.

ashleygwilliams: correct- i'm fixing it rn

ashleygwilliams: basically the idea is that we need to check that we didn't already namespace the user content id during the marky parse

ashleygwilliams:
oh heads up standard breaks because of a thing we fixed for 10.0.0 otherwise it's good to go, so don't be worried by the breaking test, it's just (the latest version of)standard and i dont think it's worth trying to fix for a patch release

ashleygwilliams:
whew actually just cherry-picked the standard fixes so travis should be happy now. good call @revin

revin:
This doesn't need to get merged back into anything, just published from here. Closing, and thanks @ashleygwilliams :)

PR #367, opened 2017-05-16
Automatically escape certain HTML tags

### LINKED ISSUE #363: HTML-wrapped code blocks contents are parsed
GitHub's Markdown parser doesn't seem to parse backticks-contained text even when put inside of HTML; the following Markdown:
```
Some content

<div>`<style>`</div>

More content - invisible.
```
will render just fine while marky-markdown cuts everything off after the `<style>` part. This broke readme of Webpack's `style-loader`; see https://www.npmjs.com/package/style-loader & https://github.com/webpack-contrib/style-loader/pull/227.

revin:
Here's a markdown-it plugin that will automatically escape `<iframe>`, `<script>`, `<style>`, `<textarea>`, and `<title>` tags in embedded HTML blocks, the way GH does.

fixes #363


revin:
Currently a WIP; needs to have [a few questions](https://github.com/npm/marky-markdown/issues/363#issuecomment-301577964) answered:

>* Our tests explicitly check that `<script>` is allowed when we're executed with `sanitize: false`. Nothing magic about `<script>` in particular; it's just an example of something that would be normally stripped out by the sanitizer. So I'm thinking maybe since turning the sanitizer off is, in a way, opting out of 100% strict GH compat, what about skipping this HTML escaping process in the case of `sanitize: false`?
>* The sanitizer is configured to strip iframes unless they're pointing to youtube URLs. IIRC we still need that capability because the npm docs have embedded YT vids. Is that still the case? It looks like GH compat is to always escape `<iframe>` tags in HTML blocks no matter what the `src` points to. Should we implement the GH version, and allow for the {YT-only, unescaped} version via some combination of options?

PR #369, opened 2017-05-18
Support shortcut github paths in URL rewrites

### LINKED ISSUE #368: relative to absolut path conversion not working with repository shortcut syntax
The conversion from a relative path e.g 

`` 

to an absolute path e.g.

`` 

is not working with a `package.json` that uses the [npm repository shortcut syntax](https://docs.npmjs.com/files/package.json#repository) 

e.g.

```json
{
  "name": "wahlberg",
  "repository": "mark/wahlberg"
}
```

it only works with

```json
{
  "name": "wahlberg",
  "repository": {
      "type": "git",
      "url": "https://github.com/mark/wahlberg"
    }
}
```

many packages out there, have links or images in their readme, that are not working or not displayed correctly on npmjs.com, but they are working on their github repository. we discussed this issue already on SamVerschueren/listr#61, because [listr](https://www.npmjs.com/package/listr) makes use of the shortcut syntax for the repository in their `package.json`.

but since `marky-markdown` is the thing that parses package READMEs on [http://www.npmjs.com](http://www.npmjs.com), it should work also with the shortcut syntax since it is in the [official docs from npm](https://docs.npmjs.com/files/package.json#repository), right?

I ran the tests from `./test/repo-github.js` with `"repository": "name/repo"`, `"repository": "github:name/repo"`, and 

```json
{
  "name": "wahlberg",
  "repository": {
      "type": "git",
      "url": "https://github.com/mark/wahlberg"
    }
}
```

and it is only working with the last one.
 
Possible solution if the shortcut syntax should be supported:

on `line 52` in `./lib/plugin/github.js` you get the url from `opts.package.repository.url` but it should also support `opts.package.repository` if `opts.package.repository.url` is `undefined`.

revin:
Fixes #368 by falling back to `opts.package.repository` if `opts.package.repository.url` is falsy.

### DIFF
--- a/lib/plugin/github.js
+++ b/lib/plugin/github.js
@@ -49,7 +49,7 @@ module.exports = function (md, opts) {
   if (!opts.package) return
   if (!opts.package.repository) return
 
-  var repo = gh(opts.package.repository.url)
+  var repo = gh(opts.package.repository)
 
   if (!repo) return
 
--- a/lib/sanitize.js
+++ b/lib/sanitize.js
@@ -130,6 +130,6 @@ function prefixHTMLids (tagName, attribs) {
   return {tagName, attribs}
 }
 
-function isAlreadyPrefixed(id, prefix) {
+function isAlreadyPrefixed (id, prefix) {
   return id.includes(prefix) && id.length > prefix.length
 }
--- a/test/repo-github.js
+++ b/test/repo-github.js
@@ -7,6 +7,7 @@ var cheerio = require('cheerio')
 
 describe('when package repo is on github', function () {
   var $
+  var $short
   var pkg = {
     name: 'wahlberg',
     repository: {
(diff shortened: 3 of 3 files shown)


revin:
For a parser like this, you can argue that any change in rendered output is a semver major, but I'm calling it a patch (by omitting the semver-major label from this PR). Doesn't really matter anyway, since the next version will be 10.0.

revin:
Changing this, because github-url-to-object [explicitly checks](https://github.com/zeke/github-url-to-object/blob/272793ef9cab8df426718ad15589fcce6b61fdad/index.js#L12-L14) that the parameter we pass is an object with a `.url` property, and uses that if it exists, or just treats the passed parameter as a string. So we don't need to do the fallback here, and can just pass the package.json `repository` directly. 💃

PR #370, opened 2017-05-26
chore: working on drafting a 10.x release

bcoe:
working on drafting the CHANGELOG for a 10.x release -- can't wait to try out some of @revin's changes that bring us closer to GitHub flavored markdown on npmjs.com; the impetus partially being that @zkat and myself are bumping into some rendering issues on our own READMEs.

How would folks feel about moving to http://conventionalcommits.org/ for future releases, so that CHANGELOG generation is less tedious.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,28 @@
-# 10.0.0 (2017-06-26)
+# 10.0.0 (2017-05-26)
+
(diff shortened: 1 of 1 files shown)


bcoe:
@ashleygwilliams having chatted a bit with @revin I say we just release the whole thing 😄

zkat:
This pleases the npm hivemind 🐝

PR #376, opened 2017-06-02
Allow <h1 align="center"> like GitHub

jaswrks:
This adds support for center-aligned H1 tags when building a centered heading for a README, which is possible at GitHub, but comes out weird at NPM. Particularly when repo shields/badges are centered, or a header image appears right underneath it. These are things that _are_ centered at NPM, while the heading itself is not — this fixes that.

### DIFF
--- a/lib/sanitize.js
+++ b/lib/sanitize.js
@@ -58,7 +58,7 @@ sanitizer.config = {
     ul: ['contains-task-list']
   },
   allowedAttributes: {
-    h1: ['id'],
+    h1: ['id', 'align'],
     h2: ['id'],
     h3: ['id'],
     h4: ['id'],


revin:
Hey @jaswrks thanks for the PR; nice find! 👍🏼

I'm guessing (and judging from a quick and dirty test in this comment box) they allow this for h1/h2/h3/h4/h5/h6. Would you care to add support for those as well?

This is also the kind of thing we like to create tests for, if only as implicit documentation regarding what GFM "is", if that makes sense. Tests for these would probably belong in [test/sanitize.js](https://github.com/npm/marky-markdown/blob/master/test/sanitize.js). Would you be up for doing the unit test implementation as well?

If not, or if you'd like any sort of guidance, just let me know; I'm happy to take this the rest of the way if you don't feel like doing it. Either way, thanks for the bug report and the fix! ✨🐬✨

thiamsantos:
Any news on this @jaswrks?

thiamsantos:
I really want to see it in production. 

@jaswrks do you have plans to finish this PR? Because, if you don't mind, I'm willing to finish it.

vutran:
Would love for this to be supported for the other heading tags as well. Any updates on getting this merged in?

jaswrks:
Sorry, got busy and missed these updates. It looks like my suggestion was [merged into master already](https://github.com/jaswrks/marky-markdown/blob/cf7ca92c465aa70c848a292ff8f0db354247c374/lib/sanitize.js#L63-L68), so this can be closed. Closing now. Thanks for the review! 😄

PR #382, opened 2017-06-04
Fix infinite loop in abstemious HTML plugin

### LINKED ISSUE #381: marky-markdown can go into an infinite loop rendering
we found a README in production that forces marky-markdown into an infinite rendering loop in v10:

https://gist.github.com/bcoe/d03e7002657a83ae6a475b492cf10539

we've downgraded to v9, and the file is able to be rendered; any thoughts @revin?

revin:
Found the issue causing an infinite loop in long documents with many consecutive HTML blocks; the rule wasn't doing an early exit when invoked in silent mode (which invocation happens when rules call other rules and only want to know if they match, without mutating the parsing state).

This also drastically speeds up rendering said documents, which is nice. I didn't make any official timing comparisons, but while hunting this down I instrumented the plugin with a silly level of logging output, and the test runs had on the order of 50x fewer lines in the logging because this fixes what ends up being a silly amount of recursion. Bonus!

Fixes #381

### DIFF
--- a/test/markdown.js
+++ b/test/markdown.js
@@ -4,6 +4,7 @@ var assert = require('assert')
 var marky = require('..')
 var fixtures = require('./fixtures')
 var cheerio = require('cheerio')
+var tripwire = require('tripwire')
 
 describe('markdown processing', function () {
   var $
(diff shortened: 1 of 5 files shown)


revin:
@bcoe I think this is ready to go. Care to verify?

test/markdown.js
```diff
@@ -356,5 +357,28 @@ describe('markdown processing', function () {
     it('processes a table', function () {
       assert.equal($('table').length, 1)
     })
+
+    it('properly renders long documents containing many consecutive HTML blocks', function (done) {
+      // the failure mode here is an infinite loop that mocha can't interfere
+      // with, so we need to set up an environment that can do it
+      //
+      function handleException (e) {
```
bcoe: aha! I better understand what the issue was now -- you couldn't even get timeouts?

bcoe:
This is great; I'm going to submit a pull request that adds support for the debugging build step, which we can try out on the next release.

test/markdown.js
revin: Yeah it just sat there running forever. Easily replicated with a `while (1) {}` loop 😕

Issue #391, opened 2017-06-08
No longer need abstemious HTML block parsing?


revin:
The reason we have the abstemious-html-block parsing plugin is because per #317 GitHub used to allow certain markdown constructs to interrupt HTML blocks, unlike standard CommonMark, which requires a blank line in between the HTML block and any subsequent markdown content.

In tracking down and testing #388, #389, & #390, I discovered that apparently GitHub has discontinued this practice. See [this gist](https://gist.github.com/revin/1d3ffd5061bb70c09a8e243c0807e4c0/3978ea5e4d40ad8cd28200dc1acc5ffe88a02e84) confirming that the markdown content following each of the images is no longer parsed as markdown.

I think we can get rid of our plugin and use markdown-it's normal HTML block parsing logic now. Thoughts, @bcoe & @ashleygwilliams?

bcoe:
@revin 👍 I approve of this plan; feels like it might correct a few bugs folks are seeing.

Issue #389, opened 2017-06-08
extraneous header tags when using horizontal rule


kentor:
I think this is a bug:

markdown source:
```
<p></p>

---

<p></p>
```

Preview:

<p></p>

---

<p></p>

Github produces this html (as expected):
```html
<p></p>
<hr>
<p></p>
```

marky-markdown (10.1.4):
```html
<p></p>
<h2></h2>
<hr>
<p></p>
```

Notice the extra `<h2>` tags.

Gets weirder if I have more stuff:

```markdown
<p>
  <a href="">asdfas</a>
</p>

---

<p></p>
```

produces

```html
<h2></h2>
<p>
  <a href="">asdfas</a>
</p>
<h2></h2>
<hr>
<p></p>
```

revin:
Hey @kentor yep, you're right, this is indeed a bug. Thanks! I'll take a look.

revin:
OK, this is fixed in 11.0.0, just released. You can verify it on https://revin.github.io/marky-markdown/ if you like, while we wait for it to be deployed to the npm site.

kentor:
Thanks for the quick fix. Really appreciate it.

PR #392, opened 2017-06-09
Remove abstemious HTML parsing behavior

### LINKED ISSUE #388: comments at top of markdown document can eat the entire document
a document starting with:

```
<!---

my comment at the top of the .md file

--->
```

results in an empty document being rendered.

revin:
As mentioned in #391, GitHub has moved closer to CommonMark HTML block parsing in that they now require HTML blocks to be terminated by a blank line. That makes our plugin that copied their old behavior obsolete, so this PR removes it, and all of the tests that verified the nonstandard behavior. This ends up fixing lots of new quirks caused by the ambiguity of the older parsing rules (which is presumably why CommonMark specifies this in the first place).

Fixes #388 #389 #391


revin:
@bcoe would you call this a major version bump, or a patch bump? In general it feels fix-y, but the 10.0 release notes specifically called this out as backwards-incompatible behavior, and now that's reverted as we chase GH

bcoe:
@revin version #s are cheap, I'd be consistent and just call it a major 😄

Issue #390, opened 2017-06-08
list items containing square brackets with no subsequent parens just disappear


SheetJSDev:
Using 10.1.4 (the latest version on NPM right now) against https://github.com/SheetJS/js-xlsx/blob/master/docbits/98_reference.md:

```markdown
 - [MS-XLSX]: Excel (.xlsx) Extensions to the Office Open XML SpreadsheetML File Format
 - [MS-OE376]: Office Implementation Information for ECMA-376 Standards Support
```

The github preview shows both lines:

<img width="694" alt="screen shot 2017-06-08 at 11 44 25" src="https://user-images.githubusercontent.com/6070939/26937618-d9c3fce6-4c3f-11e7-821a-0b925d30698c.png">

However the generated HTML shows a blank line for the second case:

```html
<li>[MS-XLSX]: Excel (.xlsx) Extensions to the Office Open XML SpreadsheetML File Format</li>
<li></li>
```

If you want to see this in the NPM website, the [xlsx README.md](https://www.npmjs.com/package/xlsx#references) is missing the same set of entries:

<img width="594" alt="screen shot 2017-06-08 at 11 46 20" src="https://user-images.githubusercontent.com/6070939/26937714-2190acb8-4c40-11e7-8f8e-3747327406b7.png">

Interestingly, the lines with square brackets and with parentheses are preserved, even though they do not correspond to a link.

revin:
Hi @SheetJSDev, very interesting. I can't think off the top of my head what might cause this, but we'll definitely get to the bottom of it.

SheetJSDev:
This is still an issue in `11.2.0`.  FWIW the workaround of escaping the opening square brackets appears to work.

daphee:
I investigated a little and found the source of the bug: the `override-link-destination-parser.js` plugin.

This plugin alters the part of `markdown-it` that parses the *target* portion of a link to allow spaces in links.

The problem is that CommonMark includes syntax to apply a label to a link in order to be able to replace the full link by just the label in other parts of the text. (e.g. see this [example](http://spec.commonmark.org/0.28/#example-168))

This syntax is available in lists as well, apparently. So even CommonMark (and GH) suffer from same bug but only for lists with one word after the `[...]` part.

 - [MS-XLSX]: Excel 
 - [MS-OE376]: Office

[...]

PR #394, opened 2017-06-09
Remove all obsolete GFM quirks handling

### LINKED ISSUE #393: Review our CommonMark deviations
GitHub seems to have quietly updated a lot of their parsing behavior. While looking into #390 I discovered that they no longer support spaces in image/link URLs or the so-called (by us) "relaxed" link reference specifications (where the link reference URL was allowed to appear without a blank line between it and the preceding paragraph).

There might be more, so we should review our plugins one by one to see if we still need them.

revin:
I can scarcely believe what I'm seeing, but since GitHub now appears to be doing CommonMark compatible parsing, that means we need to remove all of the customizations we so painstakingly assembled over the past few years 😳

CommonMark-compatible processing sometimes results in drastically different output, so I would _think_ GH might be using different renderers depending on when a README was last pushed; however, I haven't checked into this at all.

The first attempt at this PR here just removes all non-CommonMark parsing rules and their tests. Is this what we want? Or do we want the flexibility to render with old-style GitHub rendering (in which case, we'd want to keep everything, but hide it behind an option)?

Fixes #393.


revin:
Turns out this fixes #390, which is nice.

The main thing that gives me pause here is that I don't have any definite confirmation that the new CommonMark behaviors I'm seeing are permanent. GH's post from a few months ago suggested that this day would come, but I don't know if I can be sure this is forever yet.

bcoe:
@revin did you come to any decision on this issue? I'm almost tempted to keep our exiting behavior until it collides with future work we're doing, then we can take a step back and decide whether it's worth keeping custom behavior (if it has now deviated from GitHub's parser).

revin:
@bcoe Well, we do have an open issue now (#390, as above) that's a direct result of GFM getting back in line with CommonMark, so our custom parsing/rendering behaviors are pretty surely unnecessary right now (and since they deviate from GFM, we traditionally consider that a bug on our side). We'll still be wanting to do marky-specific things like rewriting URLs so image references work and things like that, but for actual markdown parsing, I'm all in on being as close to what GFM currently does as possible.

PR #396, opened 2017-06-12
Add support for JSX syntax highlighting in code blocks

### LINKED ISSUE #395: JSX language support
As noted [here](https://github.com/facebook/prop-types/pull/78), jsx-annoted code blocks which are formatted correctly on github do not use syntax highlighting on npmjs.com.

revin:
This will add support for highlighting JSX code blocks, once [highlights#55](https://github.com/atom/highlights/pull/55) is merged upstream.

Fixes #395

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ notifications:
   email: false
 
 node_js:
-  - 4
   - 6
   - 8
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ notifications:
   email: false
 
 node_js:
-  - 6
+  - 6.9.5
   - 8
 
 before_install:
--- a/lib/plugin/language-alias.js
+++ b/lib/plugin/language-alias.js
@@ -28,5 +28,6 @@ var plugin = module.exports = function (md, opts) {
 }
 
 var languageAliases = plugin.languageAliases = {
-  bash: 'sh'
+  bash: 'sh',
+  typescript: 'ts'
 }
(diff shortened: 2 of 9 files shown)


revin:
Marking as WIP for now since this is waiting on getting a version number from the `highlights` package.

zeke:
> waiting on getting a version number from the `highlights` package.

Can you elaborate on what this means? Maybe I can help unblock.

revin:
@zeke `highlights` doesn't support JSX yet, so I submitted a PR for it: https://github.com/atom/highlights/pull/55

Once that's merged, I can get the version number they publish, and fix up our dependencies here to match.

bcoe:
@zeke @revin I've published `highlights@next`, would love you to take it for a spin and make sure it works as expected.

It would be good to also add a test for TypeScript, since we also landed this feature; once you're happy @revin I can work with you sometime soon to role the renderer out to npm.

revin:
Turns out (which I forgot) that upgrading to highlights@^3 ends up breaking on node 4 because a transitive dependency uses JS `Proxy`. So it looks like we're stuck for now unless we decide to retire node 4 support.

@bcoe @ashleygwilliams thoughts?

revin:
Looks like node 4 is EOL in April 2018. Just IMHO, personally I feel like JSX and TypeScript support are probably in more demand than node 4 is, over the next 9 months. Only a hunch though, not supported by data.

bcoe:
the rendering service we're running at npm uses Node 6; I think we should land this, but bump the major -- adding support for JSX is huge.

bcoe:
(will do so once tests pass).

Issue #407, opened 2017-08-20
CLI help and version flags


SheetJSDev:
```
$ marky-markdown -h
/usr/local/lib/node_modules/marky-markdown/bin/marky-markdown.js:14
  if (err) throw err
           ^

Error: ENOENT: no such file or directory, open '/private/tmp/-h'


```

There's a help message that only shows up with no arguments:

```
$ marky-markdown 
Usage:

marky-markdown some.md > some.html
```

I'd suggest supporting a few flags:
- `-h` or `--help` to display the help message (maybe it can be collapsed to one line like `Usage: marky-markdown some.md > some.html`)
- `-V` or `--version` to display the version number (which would be available from the bin script as `require('../marky.json').version`)

revin:
Good ideas. I find myself wanting better CLI options all the time, personally.

PR #408, opened 2017-08-23
Add some CLI options

Announcement:
-v now prints the current version number.
-h now display a simple help message.

Implements feature request in issue #407 raised by @SheetJSDev

### DIFF
--- a/bin/marky-markdown.js
+++ b/bin/marky-markdown.js
@@ -2,15 +2,47 @@
 var fs = require('fs')
 var path = require('path')
 var marky = require('..')
+var information = require('../marky.json')
 
-if (process.argv.length < 3) {
-  console.log('Usage:\n\nmarky-markdown some.md > some.html')
+var executable = process.argv[0]
+var script = process.argv[1]
+var parameters = process.argv.slice(2)
+
+function printHelp () {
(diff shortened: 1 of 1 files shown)


revin:
Hi @Announcement thank you, this is awesome 🎉 🎉 🎉 

The Travis build is failing because of the `standard` formatter stuff complaining about a couple of unused variables. Do you think that's something you'd be interested in taking care of? I can totally take it over too if you'd rather do that; just say the word.

✨ Thanks so much! ✨

bcoe:
@revin @Announcement would you be open to pulling in `yargs` for this? it gives nice pretty output, and will scale better as we add more configuration options.

revin:
@bcoe 100% in favor of yargs; I'd love to have better CLI configurability, so whatever makes that easiest

nextgenthemes:
So this has no CLI options at all. Is there a config file it reads? No docs, nothing. Hours of searching for something as simple as running some super basic markdown to html in npm scripts with a modern lib like markdown-it and landed here because it does not seem to have a CLI ...

4 years and not even the most basic switches get pulled. Official NPM thing??? I basically just want it NOT to remove classes from HTML inside the markdown. Any package suggestions?

PR #411, opened 2017-09-19
feat(dep): bump oniguruma

ashleygwilliams:
oniguruma released a minor version that broke node 4 support. version 7 returns node 4 support. this PR bumps onig to that version.


bcoe:
@ashleygwilliams whoops belated, looks great to me.

PR #413, opened 2017-09-21
Bump sanitize-html version

styfle:
It looks like sanitize-html is outdated and has some security vulnerabilities. 
This PR bumps that package to the latest version.


revin:
Hi @styfle nice, thanks for having our backs! 👍

revin:
Released 11.3.1. https://www.npmjs.com/package/marky-markdown

Issue #410, opened 2017-09-05
Consider removing issuestats.com badge


Charliekenney23:
The issue-stats badges in the readme almost never loads, as [issue-stats](http://issuestats.com) has been having trouble with requests lately. The buttons should be removed or substituted with a badge from a working provider, like: [shields](https://shields.io).

Here is one for this repo:
[![GitHub issues](https://img.shields.io/github/issues/npm/marky-markdown.svg)]()

revin:
I have no opinion on badge providers, so I'd say anything that performs better (and doesn't come with any sort of legal baggage) is a win. 👍

binhonglee:
Hi, I just made a pull requests #414 in response to this issue. Hope that helps!

revin:
Yes indeed. It's published in 11.3.2. Thanks again!

PR #414, opened 2017-09-28
Addressing issue #410 and replaced issuestats badges with shields

binhonglee:


revin:
woo! thanks @binhonglee, love it  🎉 🌮 🐬

PR #416, opened 2017-10-06
Process all image/link URLs during github rewrites

revin:
The github URL-rewriting markdown-it plugin was accidentally only doing a single URL replacement per `html_block`; this change recursively processes every occurrence instead.


bcoe:
@revin 👋 I haven't had a chance to dig into this yet, but I'd be interested to see if we can replace the regex with `//g` and avoid the recursive step -- I bet we can figure something out.

bcoe:
as an alternative, if we can't, we could switch the recursive step to a stack structure and at least eliminate the chance of a stack overflow.

PR #417, opened 2017-10-13
Webpack support

DimitarTachev:
Wrapped the _require.resolve_ with an eval in order to avoid getting a number instead of the file path using Webpack. Check [https://github.com/webpack/webpack/issues/1554#issuecomment-158679844](https://github.com/webpack/webpack/issues/1554#issuecomment-158679844)

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -41,7 +41,9 @@ if (typeof process.browser === 'undefined') {
 
   languages.forEach(function (language) {
     highlighter.requireGrammarsSync({
-      modulePath: require.resolve(language + '/package.json')
+      // the eval is a workaround for a Webpack issue with resolving numbers instead of paths
+      // check https://github.com/webpack/webpack/issues/1554#issuecomment-158679844
+      modulePath: eval("require.resolve(language + '/package.json')")
     })
   })
 


bcoe:
@DimitarTachev calling `eval` always sets off alarm bells for me; my concern would be we eventually make `languages` configurable, as an example, and then have an attack vector for users injecting code.

To me this feels a bit like this is a bug with `webpack` that should be addresses -- is it that the resolve isn't working in a browser context, or where in the build process does this error happen?

DimitarTachev:
Hi @bcoe,

Thanks for getting into this. An exception is thrown runtime (node.js process) in [this highlightjs method](https://github.com/atom/highlights/blob/7874ca7bba5c8e436a6d996f0d4cdb4a04dae6f4/src/highlights.coffee#L76) as its expecting a folder path, not a number. The problem is that in the webpack context, the resolve method returns a number. Wrapping it with an 'eval' statement, it is evaluated in the nodejs context which returns a file path and everything is working as expected.

I knew that most probably you won't merge this as "eval is evil" but I wasn't able to find any other solutions. We will just use our fork as it looks safe till the eval content is always a constant.

PR #419, opened 2017-10-25
[WIP] Add formatting based on lang for html_blocks

### LINKED ISSUE #402: Support syntax highlighting in <pre>
_Originally reported at https://github.com/npm/www/issues/207._

-------

In GitHub Flavoured Markdown use of `` ```javascript`` renders the same as `<pre lang="javascript">`. Same HTML output, same syntax highlighting and other styling. Use of `<pre>` is often preferred when compatibility is desired with other Markdown parsers (such as Doxygen, Phabricator, GitBlit, GitWeb etc.), especially when the code is published from a repository outside GitHub (but possibly still mirrored to GitHub).

Many @wikimedia projects are maintained in Phabricator (with its own Markdown-like parser for readme files), and documentation generated by Doxygen (with its own Markdown-like parser to create the landing page from the readme file).

It seems marky-markdown (used on npmjs.com) treats these two syntaxes differently.

| [`` ```javascript`` on GitHub](https://github.com/patrislav/rummy.js) | [`` ```javascript`` on npmjs.com](https://www.npmjs.com/package/rummy)
|--|--
| screen shot | screen shot

| [`<pre lang="javascript">` on GitHub](https://github.com/wikimedia/oojs#oojs) | [`<pre lang="javascript">` on npmjs.com](https://www.npmjs.com/package/oojs)
|--|--
| screen shot | screen shot

rendering of the above tables

Notice how the `<pre>` element rendering at <https://www.npmjs.com/package/oojs> has no background color or syntax highlighting.

It'd be cool if marky-markdown could support this for improved interoperability (given that `<pre>` is supported by all, and also falls back nicely, whereas the tripple-backtick approach is mostly GFM-only. ReMarkup is the only other one I know of that supports tripple-backticks, but unfortunately it uses an incompatible syntax for specifying the language (it requires `` ```lang=javascript`` insteaed, which GFM doesn't support), which means the only common way to both is to use `<pre>`. And Doxygen/GitWeb/GitBlit etc. only take `<pre>` anyway.

karanjthakkar:
@revin The existing README highlighting for oojs: https://npm.im/oojs
The fixed highlighting after this change: https://auspicious-thread.surge.sh

Approach:
1. Add a plugin which overrides the html_block renderer
2. If a pre tag is available with, it checks if the lang is javascript/bash. (Couldnt find a more comprehensive list of supported langs but new ones can be added without issues)
3. Extract content from the html block
4. Replace token content with this
5. calls the parent html_block renderer and then uses the code from https://github.com/npm/marky-markdown/blob/master/lib/plugin/code-wrap.js to wrap this generated html into a code highlighting wrapper

Overall this seems to fix the issue. I'd like to get your feedback on whether I solved it in the right way. Also, I'm a little confused as to how I should go about adding tests for this. Any help/advice would be appreciated.

---
Fixes #402

### DIFF
--- a/lib/plugin/html-block.js
+++ b/lib/plugin/html-block.js
@@ -1,0 +1,38 @@
+//https://regex101.com/r/cS4Ly1/3/
+var EXTRACT_LANG_REGEX = /<pre(?:\s[^<|>]*?)lang=(?:'|")(\w+)(?:'|")/
+//https://regex101.com/r/cS4Ly1/2
+var EXTRACT_CONTENT_REGEX = /<pre(?:\s[^<|>]*?)>((.|\n)*)<\/pre>/
+var langMap = {
+  'javascript': 'js',
+  'bash': 'sh'
+}
+
+var plugin = module.exports = function(md, opts) {
+  var originalBlockRule = md.renderer.rules.html_block
+  md.renderer.rules.html_block = function(tokens, idx, options, env, self) {
+    var output = originalBlockRule.call(this, tokens, idx, options, env, self)
+    if (tokens[idx].type === 'html_block') {
+      var langMatches = tokens[idx].content.match(EXTRACT_LANG_REGEX)
+      if (langMatches[1] === 'javascript' || langMatches[1] === 'bash') {
+        var langName = langMap[langMatches[1]]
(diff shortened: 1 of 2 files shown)


karanjthakkar:
Bump.

PR #422, opened 2017-10-30
YouTube embed handling

daphee:
Like discussed in #274 this "deprecates" the youtube plugin. 

- Youtube plugin disabled by default, add option to enable it
- Sanitizer allows YouTube iframes only if the option is turned on
- Altered youtube test and turned the option on there
- Altered sanitizer tests to check that by all iframes are removed and only if the option is turned on the youtube iframes are let through.

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,7 +14,8 @@ var defaultOptions = {
   debug: false,
   package: null,
   headingAnchorClass: 'anchor',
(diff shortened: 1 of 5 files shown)


bcoe:
@daphee if I understand correctly, this will switch from Youtube being enabled by default to be enabled when a flag is set?

PR #426, opened 2018-01-17
feature: add rel=nofollow to links if the `nofollow` option is set

ceejbot:
- Added a plugin that adds rel=nofollow to links and does no more.
- Added an option to control if the plugin is loaded, defaulted to true, and documented the option.
- Wrote tests to verify that the rel attr is added, and that the option is respected.
- Added `rel` to the allowed list of attributes in the sanitizer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ The default options are as follows:
 ```js
 {
   sanitize: true,               // remove script tags and stuff
+  nofollow: true,               // add rel=nofollow to all links
   linkify: true,                // turn orphan URLs into hyperlinks
   highlightSyntax: true,        // run highlights on fenced code blocks
   prefixHeadingIds: true,       // prevent DOM id collisions
--- a/index.js
+++ b/index.js
@@ -63,7 +63,7 @@ var marky = module.exports = function (markdown, options) {
 }
 
 marky.parsePackageDescription = function (description) {
-  return sanitize(render.renderPackageDescription(description))
+  return sanitize(render.renderPackageDescription(description), defaultOptions)
 }
 
 marky.getParser = function (options) {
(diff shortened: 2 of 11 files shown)


isaacs:
This works for `[markdown](links)` but plain old `<a>` tags don't get munged.

A patch that adds support for `<a>` html tags as well, not sure if it's the most elegant approach, but seems to work.  https://gist.github.com/isaacs/16c0595f48bbaa8d44898112ff112630

revin:
Weird; what happened to the CI build? Manually canceled?

isaacs:
@revin That's correct.  I think in that case, you've explicitly said that you want to add `rel=nofollow` to markdown links, but _don't_ want to do any sanitization of inline html.  So, any inline html should be untouched (because we're not sanitizing), and thus, you're getting exactly what you asked for.  Note that you'd also be allowing `<a onclick=alert(/xss/)>`.

I think that cases where you're using nofollow on _trusted_ code would be kind of unusual anyway, no?

revin:
@isaacs yeah I'll buy that. I don't think we've ever had a particularly well defined policy on whether or not we treat markdown content and HTML content as having different semantics around their processing. I'd be very interested in just leaving raw HTML alone and being _only_ a markdown processor, although I'm like 95% sure that would break real READMEs in the wild.

revin:
Guess now we gotta figure out how to fix the CI build.

revin:
(labeling this as semver-major since the new option is defaulting to `true`)

revin:
Yeah they only seem to fail on node 6, which type of problem historically in this project indicates an upstream dependency dropping support (read: using newer JS features unsupported in the older node) somewhere along the line

PR #435, opened 2018-12-04
feat: handle subdirectories when building relative link URLs

ajafff:
This fixes an issue with relative links in packages maintained in a monorepo (described at https://npm.community/t/relative-link-in-readme-of-monorepo/3806)

You basically set `repository.url` in `package.json` to the subdirectory that contains the sources of the package. In my case that's https://github.com/fimbullinter/wotan/tree/master/packages/mimir
All relative links are relative to that URL. Absolute links are relative to the repository root.

Previously it would always append `/blob/HEAD/<link>` after the full URL.

### DIFF
--- a/lib/plugin/github.js
+++ b/lib/plugin/github.js
@@ -13,7 +13,16 @@ function buildImageUrl (repository, url) {
 }
 
 function buildLinkUrl (repository, url) {
-  return repository.https_url + path.join('/blob/' + DEFAULT_REF + '/', url.href)
+  // preserve prefix if repository url already contains /tree/<rev> or /blob/<rev>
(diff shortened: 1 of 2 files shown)


ajafff:
CI failures look unrelated to the changes in this PR

ajafff:
This is now supported using `repository.directory` in package.json

johnemau:
> This is now supported using repository.directory in package.json

Although [@ajafff's comment][1] implies this issue was fixed by setting `respositoy.directoy` in the `package.json` I do not see any implementation and the issue is still reproducing for me.

I see the examples provided before are now using fully qualified URLs to the github resource as a workaround. 

[1]: https://github.com/npm/marky-markdown/pull/435#issuecomment-493590244

Issue #442, opened 2019-08-16
Support newer versions of Node


redbugz:
See #441 

Getting compile errors in Node 12 due to out of date dependencies, specifically oniguruma.
```
└─┬ marky-markdown@12.0.0
  └─┬ highlights@3.2.0-candidate.1
    └─┬ first-mate@6.3.0
      └── oniguruma@6.2.1 
```
It appears highlights, first-mate, and oniguruma all have Node 12 compatible versions now.

mehdibeldji:
Please up !

zeke:
The [last update](https://github.com/npm/marky-markdown/commit/008509231558765695938020a376b5b2e4fd4fcc) to this project was nearly two years ago. I can't say for sure but I think it's unmaintained at this point.

If you're already using `marky-markdown` in a project and need these updates, I would suggest forking, publishing to npm under a new name, and dropping a link to the new package in this issue.

If you're just on the lookout for a decent Markdown parser with good GitHub-flavored Markdown compatibility, I would suggest [`hubdown`](https://ghub.io/hubdown), which is powered by [`remark`](http://ghub.io/remark) and `remark` plugins.

PR #443, opened 2019-10-22
feat: Support latest Node version and update dependencies 

jalalazimi:
According to #442 , I updated the packages to the latest version and fix StandardJS errors.
Now all related issues to Oniguruma has been resolved.


Flet:
Looks like the `package-lock.json` file needs to be updated in order for travis to run tests.

jalalazimi:
you're right. Also, I should fix tests as well.

PR #451, opened 2021-01-27
feat: upgrade to node 12

### LINKED ISSUE #442: Support newer versions of Node

radiantspace:
- Update `highlights` to use newer oniguruma dependency, which otherwise won't compile on Node 12.
- Remove cleanup grammar code, which doesn't work with newer `highlights`. While trying to solve some memory consumption issues, the code is actually an antipattern, which mutates the state of `highlighter` object. I kept the code for now to discuss how we want to proceed with the change.

Closes https://github.com/npm/marky-markdown/issues/442

Related to github/npm#1814﻿

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ notifications:
   email: false
 
 node_js:
-  - 8
+  - 12
 
 before_install:
   - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
--- a/lib/render.js
+++ b/lib/render.js
@@ -5,7 +5,7 @@ var emoji = require('markdown-it-emoji')
 var expandTabs = require('markdown-it-expand-tabs')
 var githubTaskList = require('markdown-it-task-lists')
 
-var cleanup = require('./cleanup')
+// var cleanup = require('./cleanup')
 var githubLinkify = require('./linkify')
 
 var codeWrap = require('./plugin/code-wrap')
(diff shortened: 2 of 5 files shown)


andrialexandrou (approved):

lib/render.js
```diff
@@ -48,7 +48,11 @@ if (typeof process.browser === 'undefined') {
 
   // cleanup generated rules in the highlighter registry if they are idle for 2000ms
   // they take a tremendous amount of memory if you process many languages in a server type environment.
-  cleanup(highlighter.registry.grammars)
+
+  // this doesn't work with new highlights version,
```
gordey4doronin, line 52: Is `3.1.4` actually newer than `3.2.0-candidate.1`? I think it's not.
Is there any problem keeping the candidate one and keep cleanup code then?

.travis.yml
```diff
@@ -4,7 +4,7 @@ notifications:
   email: false
 
 node_js:
-  - 8
+  - 12
```
gordey4doronin, line 7: I'd also add `.nvmrc` since we're touching this repo.

lib/render.js
radiantspace: Actually 3.1.4 is newer:
Screen Shot 2021-01-29 at 12 46 54

The reason for changing to newer version of highlights is to update `oniguruma` dependency, which is the actual blocker for supporting Node 12.

